PR #59, opened 2014-10-15
add no teardown option in .run

johntitus:
Added an optional boolean to .run that tells Nightmare to not teardown the instance when run is complete, so you can do some conditional control flow. 
Added a larger example, using the no-teardown option & some control flow examples.

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -70,6 +70,7 @@ function getNextPage(callback){
       callback();
     },false); //Don't teardown this instance
 }
+
 //Counts the number of links on a search results page.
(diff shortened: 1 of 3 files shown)


Oted:
What is the status of this feature? I find it very, very useful. 
Can we please have a merge to master if there are no objections?

samuelmeireles:
+1 much wanted!

reinpk:
the underlying problem will be solved in v2.0 shortly by moving to generators + an explicit `.end()` method

PR #72, opened 2014-10-22
add `html(selector, parent, callback)`

atian25:
Signed-off-by: TZ <atian25@qq.com>


reinpk:
could be potentially interesting, what's the use case? closing for now since it would need a rewrite for `2.x` but happy to re-open with more context

PR #84, opened 2014-10-24
Generic MouseEvent trigger function

Max93:


reinpk:
thanks @Max93! sorry i never got to this with v1... it looks like an interesting abstraction for other events... what events were you interested in? closing for now but may bring the idea into v2 with more info here. thanks!

PR #99, opened 2014-11-03
Update `visible` method

Hermanya:
I added style check.


reinpk:
potentially interested in pulling this into `2.x` (we have a local testing now), closing but happy to re-open if you want to mirate these changes to local testing + electron!

PR #125, opened 2014-11-30
Add screenshotBase64 method

peterjosling:


reinpk:
i don't believe electron has this (for `2.x`) but it would be interesting to make a version of `.screenshot()` that just returned the buffer!

PR #137, opened 2014-12-22
Add missing semicolons. Add test PDFs to .gitignore

azurelogic:
I forgot to ignore test PDFs in my .pdf PR. Also, adds in some semicolons to be more consistent.


reinpk:
ah good call, thanks! made the changes in parallel on master

PR #153, opened 2015-02-12
Add docs badge to README

rrrene:
Hi there,

I want to propose to add this badge to the README to show off inline-documentation: [Inline docs](http://inch-ci.org/github/segmentio/nightmare)

The badge links to [Inch CI](http://inch-ci.org) and shows an evaluation by [InchJS](http://trivelop.de/inchjs), a project that tries to raise the visibility of inline-docs (early adopters include [forever](https://github.com/foreverjs/forever), [node-sass](https://github.com/sass/node-sass) and [when](https://github.com/cujojs/when)).

The idea is to motivate aspiring Node developers to dive into Open Source projects and read the code.
It's about *engagement*, because, while testing and code coverage are important, inline-docs are the humanly engaging factor in Open Source. This project is about making people less adverse to jumping into the code and see whats happening, because they are not left alone while doing so. I know that, because I put off reading other people's code way too long in my life.

Although this is "only" a passion project, I really would like to hear your thoughts, critique and suggestions. Your status page is http://inch-ci.org/github/segmentio/nightmare

What do you think?

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -1,4 +1,6 @@
 [![Build Status](https://circleci.com/gh/segmentio/nightmare.png?circle-token=dbb94336673268633371a89865e008b70ffedf6d)](https://circleci.com/gh/segmentio/nightmare)
+[![Inline docs](http://inch-ci.org/github/segmentio/nightmare.svg?branch=master)](http://inch-ci.org/github/segmentio/nightmare)
+
 Nightmare
 =========
 


reinpk:
thanks @rrrene but the page seems a bit simplistic without restructuring our inline docs to match the way it picks up headers as methods. happy to reconsider down the road, but not a priority right now, sorry!

PR #154, opened 2015-02-14
paperSize option for PDF rendering

ravisuhag:
Passing papersize option for PDF rendering


reinpk:
great idea, can we make the parameter optional with a good default though?

ravisuhag:
@reinpk  Thanks, i have setup default to what it is production now.

reinpk:
should be possible now in `2.x`: https://github.com/segmentio/nightmare#pdfpath-options

Issue #157, opened 2015-02-24
Get current page http status code


rfink:
Would be nice to have a shortcut to get the http status code of the current page.  Pull request coming shortly (test included).

reinpk:
you can now get the http status code with `.on('did-get-response-details`, ...)`...

https://github.com/segmentio/nightmare#onevent-callback

here's an example from loading `yahoo.com`:
```
[
   "https://www.yahoo.com/",
   "http://yahoo.com/",
   200,
   "GET",
   {
      "age":[
         "0"
      ],
      "cache-control":[
         "no-store, no-cache, private, max-age=0"
      ],
      "connection":[
         "keep-alive"
      ],
      "content-encoding":[
         "gzip"
      ],
      "content-type":[
         "text/html; charset=utf-8"
      ],
      "date":[
         "Sun, 25 Oct 2015 20:13:35 GMT"
      ],
      "expires":[
         "-1"
      ],
      "p3p":[
         "policyref=\"http://info.yahoo.com/w3c/p3p.xml\", CP=\"CAO DSP COR CUR ADM DEV TAI PSA PSD IVAi IVDi CONi TELo OTPi OUR DELi SAMi OTRi UNRi PUBi IND PHY ONL UNI PUR FIN COM NAV INT DEM CNT STA POL HEA PRE LOC GOV\""
      ],
      "server":[
         "ATS"
      ],
      "set-cookie":[
         "DNR=deleted; expires=Sat, 25-Oct-2014 20:13:34 GMT; path=/; domain=.www.yahoo.com",
         "DNR=deleted; expires=Sat, 25-Oct-2014 20:13:34 GMT; path=/; domain=.www.yahoo.com",
         "PH=deleted; expires=Sat, 25-Oct-2014 20:13:34 GMT; path=/; domain=.yahoo.com"
      ],
      "strict-transport-security":[
         "max-age=2592000"
      ],
      "transfer-encoding":[
         "chunked"
      ],
      "vary":[
         "Accept-Encoding"
      ],
      "via":[
         "http/1.1 ir32.fp.gq1.yahoo.com (ApacheTrafficServer)"
      ],
      "x-frame-options":[
         "DENY"
      ]
   }
]
```

PR #158, opened 2015-02-24
Adding http status code

rfink:
Addresses #157

### DIFF
--- a/lib/actions.js
+++ b/lib/actions.js
@@ -31,6 +31,7 @@ exports.use = function(plugin, done){
 
 exports.goto = function(url, done) {
   debug('.goto() url: ' + url);
+  this.url = url;
   this.page.open(url, function(status) {
     debug('.goto() page loaded: ' + status);
     setTimeout(done, 500);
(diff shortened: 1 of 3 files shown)


johntitus:
I like the idea, but this implementation is setting the status whenever the `onResourceReceived` callback fires.  That callback is "[invoked when a resource requested by the page is received](http://phantomjs.org/api/webpage/handler/on-resource-received.html)".  Every external javascript, stylesheet, etc causes this to fire.  So the status code you receive may be for something other than the URL itself.

rfink:
Right, that makes sense.  I included a check in there to make sure that the url of the response matched the url of the page requested by 'goto'

johntitus:
I see, sorry I missed that. Next question - what happens if the user sets a `resourceReceived` callback? May need to catch that an wrap it in the `exports.on` method.

rfink:
No worries, I'm happy you are keeping a close eye on it.  As far as the next issue, I'll take a look and see if I can make a recommendation.  Great point.

johntitus:
One more potential issue - If you `.goto('http://www.reddit.com')`, it gets translated into `http://www.reddit.com/` with the trailing slash. I think this would cause this method to 'miss' the response.

johntitus:
Possible way to do it, at the end of this conversation:

https://github.com/ariya/phantomjs/issues/10185

reinpk:
in `nightmare@2.x` you can listen for the `did-get-response-details` event, and inspect the arguments for the status code — there's enough information in there to pinpoint the html/url requested

PR #160, opened 2015-03-01
Pullrequest

samuelmeireles:


reinpk:
we're addressing these ideas in v2 with generators and an explicit `.end()` method, thanks!

PR #164, opened 2015-03-12
Handling Phantom crashes more gracefully

dominicbarnes:
We have a service that is running nightmare as part of it's API. If the PhantomJS process dies unexpectedly, the `phantom` lib will throw an **[uncaught exception](https://github.com/sgentle/phantomjs-node/blob/2645246eb987b163b233cc042715bd0a6a4d181f/phantom.coffee#L91-L92)**. As is the case with uncaught exceptions, this crashes the entire server. Thankfully, `phantom` does expose an [`onExit` option/hook](https://github.com/sgentle/phantomjs-node/blob/2645246eb987b163b233cc042715bd0a6a4d181f/phantom.coffee#L88-L89) that allows you to handle such cases.

As far as nightmare is concerned, this adds a new custom event. (ie: `Nightmare#on('exit', ...);`) When this handler is attached, it will be called upon with the given exit `code` and `signal`. (the signal is not always present, but in our case we saw a lot of `SIGSEGV`) If the handler is _not_ bound, then we are still forced to `throw` (so the failure isn't just lost in the abyss) but we format a pretty clear error message. (and provide the `code` and `signal` as properties on the `Error` object for later inspection.

In short, this makes no API changes, only the custom event is added. (since "exit" is not a phantom event, I have adopted it for this purpose)

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -195,6 +195,7 @@ Capture page events with the callback. You have to call `.on()` before calling `
 * `prompt` - callback(msg, defaultValue)
 * `error` - callback(msg, trace)
 * `timeout` - callback(msg) fired when a .wait() times out before condition becomes true
+* `exit` - callback(code, signal) fired when the underlying phantom process dies unexpectedly (if not handled, this will become an uncaught exception)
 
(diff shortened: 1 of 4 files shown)


dominicbarnes:
Looks like the tests that are failing here are also failing on `master`, my tests are fine though.

PR #174, opened 2015-04-27
Fix escaping of selectors in .wait(selector)

thotypous:
Fixes this error: 

```
phantom stdout: SyntaxError: DOM Exception 12: An invalid or illegal string was specified.
```

when trying to use escaped selectors (e.g. `.wait("#escaping\\:test")`)


reinpk:
lgtm thanks!

reinpk:
released with 1.8.1

Issue #175, opened 2015-04-29
Split run into setup -> execQueue -> teardown


inca:
Opening an issue to explain upcoming PR.

It would be handy to be able to work with a single PhantomJS instance multiple times (queue some things, run them, do some stuff, queue some more things, run again, etc.). For example in a typical BDD-style spec someone would want to do this:

```js
describe('My awesome page', function() {
  var nightmare = new Nightmare();

  before(function(done) {
    nightmare.setup(done);
  });

  it('should login', function(done) {
    nightmare.goto('/some/login/page').exec(done);
  });

  it('should do something stateful with a page and assert some results', function(done) {
    nightmare.click('btn--primary')
      // assert something
      .exec(done);
  });

  it('should do some more stuff', function(done) {
    nightmare
      // do something
      // assert something
      .exec(done);
  });

  after(function() {
    nightmare.teardownInstance();
  });

});
```

The funny thing is that it's super-easy to support this flow, all we have to do is:

  * extract a method which actually runs a queue from `run` (I called it `exec`)
  * make `setup` and `teardownInstance` methods public (document them, remove `@api private`)
  * (optionally) rename `teardownInstance` into `teardown`

Implementing this would also solve #36 and other similar issues. Although it does not introduce control explicit flow helpers, it becomes kinda doable with explicit `exec` calls (callback pyramid can be optimized/eliminated in numerous ways).

As for automated tests (with Mocha), this solution fits the bill ideally: async tests work like a charm and you also gain a significant performance boost by instantiating PhantomJS only once per `describe` or even per entire spec.

nylen:
A big :+1: to this!  I'm currently hacking together something similar by disabling `teardownInstance` and manually calling it when I'm ready.

inca:
@nylen Thanks! I've also included a [small test in PR](https://github.com/inca/nightmare/blob/56d3bb262d0a887f232c297ac587741c065a0c8a/test/index.js#L743) to illustrate a potential use case with `localStorage`, but it could really be anything.

reinpk:
the extra flow control here is a great idea, though we wanted to hold off and implement it using generators. v2 (#200) should get us there with the explicit `.end()` as well

PR #176, opened 2015-04-29
Split run into setup -> execQueue -> teardown

inca:
As explained in #175, I extracted the `exec` method which runs the queue on already initialized Nightmare instance. I also took the liberty of removing `@api private` from `setup` and `teardownInstance`, since these methods should now be exposed for this style of working with Nightmare.

I'd also rename `teardownInstance` into `teardown`, but I decided to leave this one up to the core team.

I'd also be happy if someone could reflect this in docs (I'm not so confident about my English).

Thanks again for an awesome tool!

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -51,7 +51,8 @@ function Nightmare (options) {
 }
 
 /**
- * Run all the queued methods.
+ * Setup new PhantomJS instance, execute all queued methods
+ * shutdown PhantomJS and finally execute a callback.
  *
  * @param {Function} callback
  */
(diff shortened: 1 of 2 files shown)


inca:
Could some of the collaborators please comment this one? Do you plan to accept this (with modifications maybe) or instance reuse is not something you plan to ever support for Nightmare?

TIA

nylen:
A big :+1: to this!  Here's what I'm currently doing instead, it's really ugly:

```js
var browsers = {};

// Do not teardown PhantomJS instances (reuse them to save time)
// Even though Nightmare#setup() is called for each Nightmare#run() call, this
// still works because Nightmare#setupInstance() checks
// this.initializingPhantomJS and only creates a new PhantomJS page, not a
// whole new instance.
// TODO this is a nasty hack that may cause other unexpected problems
Nightmare.prototype.oldTeardownInstance = Nightmare.prototype.teardownInstance;
Nightmare.prototype.teardownInstance = function() {
	require('nightmare/node_modules/debug')('nightmare')('skipping PhantomJS teardown');
};

// but do tear them down on process exit
// we specify --no-exit in the mocha.opts file, but mocha still adds a process
// 'exit' listener to set its exit code; when that listener is added, we know
// mocha is finished running
process.on( 'newListener', function( event, listener ) {
	if ( 'exit' === event ) {
		// This doesn't work
		Object.keys( browsers ).forEach( function( role ) {
			browsers[ role ].oldTeardownInstance();
		} );
		// Give everything some time to exit before forcing
		var timeout = setTimeout( function() {
			Object.keys( browsers ).forEach( function( role ) {
				try {
					browsers[ role ].phantomJS.process.kill();
					console.error( 'Killed browser process for role: %s', role );
				} catch (e) { }
			} );
			process.exit();
		}, 2000 );
	}
} );
```

inca:
@nylen Thanks for your feedback. After a couple of days I came up with a conclusion that this project is not actively maintained. FWIW we decided to switch to [navit](https://github.com/nodeca/navit) which is designed for testing right from the start.

reinpk:
we decided in v2 #200 to solve this with generators and explicit `.end()` method

PR #177, opened 2015-05-01
Adding Nightmare..drainQueue

Zirak:
It can be used when one doesn't want to kill the page once all queued actions are done.

I've already used it a few times like so:

```js
nightmare
    .stuff()
    .moreStuff()
    .setup(function () {
        nightmare.drainQueue(console.log.bind(console, 'drained'))
    });
```

It also allows queueing up more actions after the run loop, for instance from user input.

Useful enough to merge into main?

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -51,7 +51,7 @@ function Nightmare (options) {
 }
 
 /**
- * Run all the queued methods.
+ * Run all the queued methods and when done, terminate the page.
  *
  * @param {Function} callback
  */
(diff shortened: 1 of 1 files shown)


inca:
Seems like a dup of #176. Yes, this feature is highly anticipated; however, they don't merge these PRs for some obscure reason. I doubt this repo is maintained at all :confused:

reinpk:
thanks @Zirak! sorry for the delay... this pull request and others like it were influential in the work that @matthewmueller put into v2... which uses generators and an explicit `.end()` method. should solve the same problem in a slightly different way. thanks!

PR #180, opened 2015-05-10
improves resourceRequestStarted event

rawberg:
Before this change resourceRequestStarted would clobber any listeners set for resourceRequested. This change allows two callbacks to be passed to resourceRequestStarted. The first callback runs in the scope of phantom and the second runs in the scope of node. Wraps [onResourceRequested of the underlying phantom library](https://github.com/sgentle/phantomjs-node#functionality-details).


joostverdoorn:
This PR is pretty important. The current functionality seems to be faulty and I think this PR should be merged to fix that.

reinpk:
in `2.x` you should be good to go! https://github.com/segmentio/nightmare#onevent-callback

Issue #179, opened 2015-05-07
screenshot() throws "Must include file extension in `path`" when the image file path contains dots more than one


t-suwa:
screenshot() fails when the image file path includes plural dots like this:

```
c:\Users\foo_bar.DOMAIN\dev\some-product\test.e2e\evidence\case-001.png
```
Instead of using combination of `substring` and `indexOf`, we can extract the extension by regexp.

```javascript
exports.screenshot = function (path, done) {
  var formats = ['png', 'gif', 'jpeg', 'jpg', 'pdf'];
  var ext = path.match(/[^.]+$/); // use regexp to extract the extention
  if (!~formats.join(',').indexOf(ext)) {
    done(new Error('Must include file extension in `path`.'));
  }
  debug('.screenshot() saved to ' + path);
  this.page.render(path, done);
};
```

reinpk:
released with 1.8.2

PR #183, opened 2015-05-27
Fix screenshot() for paths with 2+ periods (#179)

af:
@t-suwa proposed this fix in #179, just creating a PR for easy merging.

I'm running into this problem on CircleCI. Their artifacts directories have extra periods in them, and this currently breaks any use of `screenshot()` when trying to save to them.


reinpk:
good call thanks guys!

reinpk:
released with 1.8.2

af:
Thanks!

Issue #120, opened 2014-11-15
Form Fields not targeted properly


dtothefp:
I've noticed since changes were merged in last weekend that Nightmare no longer fills out form fields properly.  You can see an example in my repo here

https://github.com/dtothefp/nightmare-timeout-test-example/tree/master

If you run the tests, generated screenshots show that form inputs do not contain the correct information.  If you switch to my branch where I use an older version of Nightmare I previously forked:

https://github.com/dtothefp/nightmare-timeout-test-example/tree/dfoxpowell/nightmare-v1.6.0

```
rm -rf node_modules
npm cache clean
npm i
```

If you previously `npm i` in the master branch.  Then run the tests again and you will see that form fields are filled out properly.  You will also see in this branch the test timeout failures I submitted previously in

https://github.com/segmentio/nightmare/issues/102

dtothefp:
@reinpk Any news on this, or should we file an issue elsewhere? I've had my Nightmare version constrained for far too long, and there have been many updates in `node phantom` module. I'm also working to switch to PhantomJS 2.0.0 which surprising worked fine if I was on Nightmare 1.6.0 but upgraded old Nightmare package.json to 0.7.2 of https://github.com/sgentle/phantomjs-node. Even the simple Optimizely signin form fails, as you can see the email and password get entered in the same input:

![screen shot 2015-03-22 at 11 49 50 pm](https://cloud.githubusercontent.com/assets/4656726/6773885/d36c2de8-d0ed-11e4-82c9-a121ef1972af.png)

dtothefp:
@reinpk I've actually pinpointed the problem to this PR https://github.com/segmentio/nightmare/pull/100/files that switches the `type` action to implement the `sendEvent` Phantom method. I'm not sure if this is a problem with `phantomjs-node`'s implementation of `sendEvent` or a problem in PhantomJS itself, but for now would you consider reverting as it doesn't seem worthwhile to have such a small change potentially effecting users?

When I get some time later in the week I can put together some tests to pinpoint the problem more acuratelly and determine if I should submit an issue with `phantomjs-node` or PhantomJS

reinpk:
yeah, that change seems to have made a lot more issues than it solved... the reason we switched it is because some fancier forms didn't respond well to the value setting method... but it was much simpler and more robust :P see #81, maybe there's a hack that sets value and then does a single `sendEvent` or something.

yeah, i'm definitely open to reverting that change and looking for a better solution.

[...]

PR #195, opened 2015-06-26
remove keypress from .type action

dtothefp:
This PR addresses https://github.com/segmentio/nightmare/issues/120 when the `keypress` event was added it broke functionality for certain forms. 

I ran tests locally and the following tests are not passing:
```js
it('should upload a file', function (done) {
  new Nightmare()
    .goto(fixture('upload'))
    .upload('input[type=file]', 'test/files/test.css')
    .click('button[type=submit]')
    .wait(1000)
    .evaluate(function () {
      return JSON.parse(document.body.querySelector('pre').innerHTML)
    }, function (files) {
      files.file.originalname.should.equal('test.css');
    })
    .run(done);
});

// results in err => Uncaught TypeError: Cannot read property 'originalname' of undefined
```

```js
it('should set authentication', function (done) {
  new Nightmare()
    .authentication('my', 'auth')
    .goto(fixture('auth'))
    .evaluate(function () {
      return JSON.parse(document.querySelector('pre').innerHTML);
    }, function (data) {
      data.should.eql({ name: 'my', pass: 'auth' });
    })
    .run(done);
});
//results in err => Error: timeout of 40000ms exceeded
```

but this seems unrelated to the PR and I'm wondering if these tests are somehow "flaky"?.

### DIFF
--- a/lib/actions.js
+++ b/lib/actions.js
@@ -222,13 +222,10 @@ exports.mouseover = function(selector, done) {
 
 exports.type = function(selector, text, done) {
   debug('.type() %s into %s', text, selector);
-  var self = this;
-  this.page.evaluate(function(selector, text){
-    document.querySelector(selector).focus();
-  }, function(){
-    self.page.sendEvent('keypress', text, null, null, 0);
-    done();
-  }, selector, text);
+  this.page.evaluate(function(selector, text) {
+    var element = document.querySelector(selector);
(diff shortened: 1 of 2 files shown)


reinpk:
cool, moving to electron in v2 which should fix this keypress issue *and* the flaky test issue ;)

PR #214, opened 2015-09-07
fix end() if no done passed as argument

xouabita:
capture d ecran 2015-09-08 a 00 11 21

When I run `make test`, it threw `Cannot read property 'apply' of undefined`.
Now it checks if done is passed as an argument.

Another fix for the if could be:

~~~js
if (done) {
  return this.run(done)
}
~~~

Because as I understand, `done` is only passed if nightmare is not running ?

### DIFF
--- a/lib/nightmare.js
+++ b/lib/nightmare.js
@@ -185,7 +185,7 @@ Nightmare.prototype._inject = function(js, done) {
 
 Nightmare.prototype.end = function(done) {
   this.ending = true;
-  if (!this.running) {
+  if (done && !this.running) {
     return this.run(done)
   }
   return this;


matthewmueller:
Ahh my bad. Basically I think the two ways you should be able to close the connection are:

Run nightmare one time

```js
Nightmare()
  .goto('http://google.com')
  .title()
  .end()
```
 Run it a bunch of times and eventually end:

```
var nightmare = Nightmare()

yield nightmare
  .goto('http://google.com')
  .title()

yield nightmare
  .goto('http://yahoo.com')
  .title()

yield nightmare.end()
```

So actually, maybe it shouldn't close the connection at all (the previous would get invoked with `.then()`). 

I think maybe it should be a no-op in run instead of conditionally running `this.run()` based on the callback.

matthewmueller:
Just tested this out, good call. This is the right solution. Thanks!

PR #217, opened 2015-09-13
implement screenshots

matthewmueller:
fixed up the `nightmare.screenshot(path)` implementation. Basically passes the image buffer back from the electron process to the node process to write the file.

Here's an example:

```js
var Nightmare = require('./')

Nightmare()
  .goto('https://google.com')
  .screenshot('screenshot.png')
  .end(function(err) {
    if (err) throw err
  })
```

### DIFF
--- a/lib/runner.js
+++ b/lib/runner.js
@@ -114,9 +114,7 @@ app.on('ready', function() {
   parent.on('screenshot', function(path) {
     // https://gist.github.com/twolfson/0d374d9d7f26eefe7d38
     win.capturePage(function handleCapture (img) {
-      fs.writeFile(path, img, function (err) {
-        parent.emit('screenshot', arguments);
-      });
+      parent.emit('screenshot', img.toPng())
     });
   });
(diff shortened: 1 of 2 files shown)


reinpk:
bloody beautiful :)

lib/runner.js
```diff
@@ -114,9 +114,7 @@ app.on('ready', function() {
   parent.on('screenshot', function(path) {
     // https://gist.github.com/twolfson/0d374d9d7f26eefe7d38
     win.capturePage(function handleCapture (img) {
-      fs.writeFile(path, img, function (err) {
-        parent.emit('screenshot', arguments);
-      });
+      parent.emit('screenshot', img.toPng())
```
matthewmueller: we can probably encode differently if it's `.jpg` vs `.png`

reinpk: ah, hmmm let's just document "only png" in the readme for now and see if it's an issue.

PR #222, opened 2015-09-15
Fire change and input events on type

pastawoua:
Changing the value of a field doesn't trigger the change and input event by default.

### DIFF
--- a/lib/actions.js
+++ b/lib/actions.js
@@ -135,15 +135,15 @@ exports.mouseover = function(selector, done) {
 exports.type = function(selector, text, done) {
   debug('.type() %s into %s', text, selector);
   this._evaluate(function (selector, text) {
-    var elem = document.querySelector(selector);
-    elem.focus();
-    elem.value = text;
+    var element = document.querySelector(selector);
+    element.focus();
+    element.value = text;
     ['change', 'input'].forEach(function(name) {
       var event = document.createEvent('HTMLEvents');
       event.initEvent(name, true, true);
-      elem.dispatchEvent(event);
+      element.dispatchEvent(event);
     });
-    elem.blur();
+    element.blur();
   }, done, selector, text);
 };
 
--- a/lib/actions.js
+++ b/lib/actions.js
@@ -138,6 +138,11 @@ exports.type = function(selector, text, done) {
     var elem = document.querySelector(selector);
     elem.focus();
     elem.value = text;
+    ['change', 'input'].forEach(function(name) {
+      var event = document.createEvent('HTMLEvents');
+      event.initEvent(name, true, true);
+      elem.dispatchEvent(event);
+    });
     elem.blur();
   }, done, selector, text);
 };
--- a/package.json
+++ b/package.json
@@ -15,7 +15,7 @@
   "dependencies": {
     "debug": "^2.2.0",
     "defaults": "^1.0.2",
-    "electron-prebuilt": "^0.32.2",
+    "electron-prebuilt": "^0.33.0",
     "enqueue": "^1.0.2",
     "function-source": "^0.1.0",
     "jsesc": "^0.5.0",
--- a/test/index.js
+++ b/test/index.js
@@ -345,7 +345,7 @@ describe('Nightmare', function () {
         });
       var stats = fs.statSync('/tmp/nightmare/test.png');
       var statsClipped = fs.statSync('/tmp/nightmare/test-clipped.png');
-      statsClipped.size.should.be.at.least(500);
+      statsClipped.size.should.be.at.least(300);
       stats.size.should.be.at.least(10*statsClipped.size);
     });
 


matthewmueller:
yah, i was worried about this, it's a much more complicated prospect to get the events working.

I think we'd want to use something like this: https://github.com/DamonOehlman/simkey if we decide to support this.

pastawoua:
Doesn't have to be very complicated, interested by the change event more than anything really.

It was on the other methods but not the type one. Do you think this can be merged in?

pastawoua:
Thanks for fixing the tests @matthewmueller

pastawoua:
@matthewmueller can this be merged in please?

rianby64:
I need to test some simple things written in Angular, and $digest doesn't see the changes. Please, merge.

pastawoua:
@matthewmueller is there an issue with the code, anything you want me to change, would be helpful to merge so I don't use my forked version in production.

fritx:
@pastawoua at least the `input` or `change` event you dispatched is missing some properties like `keyCode`, etc?

pastawoua:
keyCode is not fired on `input` or on `change` events.

fyi, casperjs implementation: https://github.com/n1k0/casperjs/blob/master/modules/clientutils.js#L899

rianby64:
@pastawoua wrote a nice solution that works for me. But if you want to follow the code style, then check this:  

```javascript
exports.type = function(selector, text, done) {
  debug('.type() %s into %s', text, selector);
  this._evaluate(function (selector, text) {
    var element = document.querySelector(selector);
    var event = document.createEvent('HTMLEvents');
    element.value = text;
    event.initEvent('change', true, true);
    element.dispatchEvent(event);
  }, done, selector, text);
};
```

So, focus() and blur() are unnecessary. At least while you're implementing the real cool things (sendKeys or simkey and a lot of more) we can use nightmare for our humble testing. Thanks

zerkalica:
Not working for react apps. Really helps this:

```js
            .type('input[name=password]', '1234')
            .evaluate(() => {
                document.querySelector('input[name=password]').dispatchEvent(new Event('input', {bubbles: true}))
            })
```

danhalliday:
@zerkalica I can confirm that `evaluate()` function works for Angular 2 forms as well.

rianby64:
@zerkalica nice that info! Have you tried to type some text in a field? I know that $digest in Angular 1.4* listens only events. So, the current version indeed types something but this is not enough. A change event should be fired. The solution given by @pastawoua is enough for me.

danhalliday:
@zerkalica Yes, I have a full end-to-end test for my Angular 2 app working using Nightmare. I work around the digest/diffing by `wait()`ing for DOM elements. It works surprisingly well (much easier than setting up Protractor, and less tightly coupled to Angular, too). Eg:

    yield nightmare
      .goto(URL)
      .wait("#login") // <form>
      .type("#login-email", EMAIL).evaluate(() => { document.querySelector("#login-email").dispatchEvent(new Event("input", { bubbles: true })); })
      .type("#login-password", PASSWORD).evaluate(() => { document.querySelector("#login-password").dispatchEvent(new Event("input", { bubbles: true })); })
      .click("#login-submit")
      .wait("#toolbar");

    var toolbar = yield nightmare
      .evaluate(() => {
        return document.querySelector("#toolbar").textContent;
      });

    expect(toolbar).to.contain(USERNAME);

rianby64:
@danhalliday Thanks a lot! I hope this kind of workarounds will be temporary. Soon we'll be able to call type(...) and go ahead with the next step.

zerkalica:
Some patch:

```js

import Nightmare from 'nightmare'
import __debug from 'debug'

const debug = __debug('nightmare:actions')

const actions = {
    input(selector, text, done) {
        debug('.input() %s into %s', text, selector)
        this._evaluate(function (selector, text) {
            var elem = document.querySelector(selector)
            elem.focus()
            elem.value = text
            elem.dispatchEvent(new Event('input', {bubbles: true}))
            elem.blur()
        }, done, selector, text)
    }
}

class MyNightmare extends Nightmare {
}
Object.keys(actions).forEach(name => {
    const fn = actions[name]
    MyNightmare.prototype[name] = function() {
        debug('queueing action "' + name + '"')
        const args = [].slice.call(arguments)
        this._queue.push([fn, args])
        return this
    }
})

const nightmare = new MyNightmare()

nightmare
.goto('login-page')
.input('input[name=password]', '1234')
```

rianby64:
@zerkalica Cool!

danhalliday:
Any progress on this?

bausmeier:
Would love to see this merged. Having to manually apply the patch at the moment.

pastawoua:
fyi, I gave up on having this merged and maintaining my forked version so we moved to http://nightwatchjs.org

rosshinkley:
I believe this was (eventually) fixed in #548.

PR #230, opened 2015-09-18
Fix electron errors casued by process not connected

davidnaas:
A non connected node process in the IPC emitter would cause electron to
break.


matthewmueller:
cool, so what would cause this situation to come up?

davidnaas:
Got this problem as soon as I started with v2.0. trying to run mocha tests locally with a react app.

PR #234, opened 2015-09-19
Resurrect .useragent api from v1, b275ff0.

jefeweisen:
Great work guys.  I switched over to the atom editor in May and have been impressed with the stability of atom-shell (now electron).  Nice choice of dependencies for browser automation!

Cheers!

### DIFF
--- a/lib/actions.js
+++ b/lib/actions.js
@@ -365,6 +365,19 @@ exports.viewport = function (width, height, done) {
 };
 
 /**
+ * Set the useragent.
+ *
+ * @param {String} useragent
+ * @param {Function} done
+ */
+
+exports.useragent = function(useragent, done) {
+  debug('.useragent() to ' + useragent);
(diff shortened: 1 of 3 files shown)


lib/actions.js
```diff
@@ -365,6 +365,19 @@ exports.viewport = function (width, height, done) {
...
 
 /**
+ * Set the useragent.
+ *
+ * @param {String} useragent
+ * @param {Function} done
+ */
+
+exports.useragent = function(useragent, done) {
+  debug('.useragent() to ' + useragent);
```
matthewmueller: i think we should still listen for `this.child.once('useragent', ...)` since the IPC communication is async even if the useragent setter isn't.

matthewmueller:
<3

reinpk:
i'll fix this up on master

reinpk:
released as `2.0.5`, thanks!

PR #237, opened 2015-09-21
Remove after and util in test/index.js

38elements:
`after` and `util` are not used in test/index.js.


reinpk:
:dancer: code deletion i love it

Issue #242, opened 2015-09-22
Page's JavaScript is not running


antpaw:
I've cloned this repo and tried `make test`, it succeeded. Then I went ahead and modified the 'navigation' fixture, (html is appended in this message), to checkout Electron's JS abilities. Looks like absolutely no JS is running, I tried to do `.click('.a')` and instead of creating a `<div>` it took me to a.html (`.title()` was no longer 'Navigation' but 'A') I tried to take a screenshot right after `goto and wait(1000)` the `div#foo` was not rendered. So I'm quite sure JS doesn't run at all, that makes v2 inferior to v1? Am I missing something?

`````html
<!DOCTYPE html>
<html>
  <head>
    <title>Navigation</title>
  </head>
  <body>
    <a class="a" href="a.html">A</a>
    <a class="b" href="b.html">B</a>
    <a class="c" href="c.html">C</a>
    <script src="http://code.jquery.com/jquery-2.1.4.min.js"></script>
    <script>
        $(document.body).append('<div id="foo">bar</div>');
        $('.a').click(function(e){
            e.preventDefault();
            setTimeout(function(){
                $(document.body).append('<div id="no">yes</div>');
            }, 2000);
            return false;
        });
    </script>
  </body>
</html>
`````

**Comment by mpangrazzi** [comment]
Looks like Electron doesn't like JS loaded from an external website. To fix your problem, simply load jQuery in this way:

 ```js
<script src="http://code.jquery.com/jquery-2.1.4.min.js" onload="window.$ = window.jQuery = module.exports;"></script>
```

Then your jQuery snippet below will be executed. You can find more details [here](https://github.com/atom/electron/issues/345).

antpaw:
Okay I will try this, the other question is: why is nightmare not giving me any errors on this one. Like `$ "undefined is not a function"`. This would make it much simpler to debug.

mpangrazzi:
Agreed. 
This weird Nightmare "behaviour" took me some time to figure how to fix it. Nightmare should have an option for reporting all Electron error or debug messages.

antpaw:
Okay i've just tried it and it looks like Electron gives the browser a package management system just like in node. So the browser has no longer to deal with global var. Instead of doing this onload hack, we can simply

`````html
	<script src="http://code.jquery.com/jquery-2.1.4.min.js"></script>
    <script>
		var $ = require('jquery');
        $(document.body).append('<div id="xxx">lolxxx</div>');
        $('.a').click(function(e){
[...]

Issue #249, opened 2015-09-25
google.com blocked in China while running test


fritx:
Could I replace it just with ~~`yahoo.com`~~`github.com`? I can use a VPN though, but not all the time..

fritx:
Oops, because of #247 , yahoo.com won't work now, looking for another popular site that works in China.

FredXue:
Po主 好嗨！

fritx:
@FredXue 嗨，我哭了

FredXue:
终于看到同胞在鼓捣nightmare O(∩_∩)O （这个issue跟nightmare没关系啊）

FredXue:
@fritx  以后遇到问题可以交流交流！

fritx:
nice  XD 毛问题 已follow

fritx:
sometimes even `<link href='http://fonts.googleapis.com/css?family=Oswald' rel='stylesheet'>` is blocked too.. it's really disappointing.. but let's just leave it for now..

PR #250, opened 2015-09-25
Use node-integration for electron, fix #247 & amd scripts loading

### LINKED ISSUE #247: $ is not defined, module ipc is not availiable
I am getting a lot of errors like $ is not defined, module ipc is not availiable, could not load jquery when i tried to login to linkedin using nightmare. Can you please tell how i can resolve this?

fritx:
~~Sorry for the extra field in package.json & a missing semicolon, but I think they could be fixed with the next bump clean~~ (cleaned) ;) fix #247 #242

### DIFF
--- a/lib/javascript.js
+++ b/lib/javascript.js
@@ -13,8 +13,8 @@ var minstache = require('minstache');
 var execute = [
   "(function javascript () {",
   "  var log = console.log;",
-  "  var ipc = require('ipc');",
-  "  var sliced = require('sliced')",
+  "  var ipc = __nightmare.ipc;",
+  "  var sliced = __nightmare.sliced;",
   "  console.log = function() {",
   "    ipc.send('log', sliced(arguments).map(String));",
   "  }",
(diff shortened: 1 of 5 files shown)


matthewmueller:
Thanks for the PR! Can you explain in a little more detail what this does to fix the problem? Some questions I have:

- Was the missing semicolon causing us to not get the errors?
- What's the purpose of preloading and disabling node-integration?

fritx:
@matthewmueller ok.

- No, forget the missing semicolon. I meant it was I that had [missed a semicolon](https://github.com/segmentio/nightmare/pull/250/files#diff-9dcc65cc4a6c14eb4f81fbe4005301d0R41) in my PR commit, but just leave it ;)
- Yes, with `node-integration` turned off:
 - electron won't expose node stuff like `module/exports` to the web
 - so we could correctly load any webpage with many scripts like jquery that depends on a `module/exports` detection
 - and with a `preload` script we can still inject anything with node access just like what we do, but only at the very beginning.

fritx:
See also: https://github.com/sindresorhus/caprine/blob/master/index.js#L39

fritx:
Let me clean my commit to be better.

fritx:
@matthewmueller once this problem solved, in #249 i'd suggest we replace `google.com` to `github.com`, and it'd get to work. ;)

matthewmueller:
thanks buddy, appreciate the research!

i'm fine with changing the tests to point to github to avoid the great firewall. open a pr and i'll accept :+1:

fritx:
cool, thanks much! @matthewmueller

PR #260, opened 2015-10-01
Update description in package.json

tscanlin:
Update description in package.json since you guys aren't using Phantom.js anymore.

Electron ftw!


reinpk:
thanks!

PR #268, opened 2015-10-05
Remove old evaluate method, fix #257

### LINKED ISSUE #257: Two exports.evaluate implementations in actions.js
There are two `exports.evaluate` implementations [#L28](https://github.com/segmentio/nightmare/blob/master/lib/actions.js#L28) and [#L312](https://github.com/segmentio/nightmare/blob/master/lib/actions.js#L312)

38elements:
#257

### DIFF
--- a/lib/actions.js
+++ b/lib/actions.js
@@ -22,32 +22,6 @@ exports.title = function(done) {
 };
 
 /**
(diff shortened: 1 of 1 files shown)


matthewmueller:
good catch

PR #273, opened 2015-10-06
Add test case for 'type'

fritx:
The test case of 'type and click' doesn't ensure 'type' works
'type and click' passes even when 'type' types nothing


fritx:
Is there any problem still on this PR?

matthewmueller:
nope! sorry about that

fritx:
Thanks ;)

Issue #203, opened 2015-08-04
Interacting with elements inside an iframe


rclai:
Is this possible?

rclai:
I had to create a `withFrame` action function that goes inside the iframe context and a `backToParentFrame` function to go back to the parent frame:

```js
exports.withFrame = function(frameSelector, fn, done){
  // `frameSelector` is either an index number or the iframe's name
  // no class names or ids unfortunately
  // As you can see, the code is identical to that of the `use` function
  debug('.withFrame():' + frameSelector);
  this.page.switchToFrame(frameSelector);
  var cache = this.queue;
  this.queue = [];
  fn(this);
  var self = this;
  this.queue = this.queue.concat(cache);
  done();
};

exports.backToParentFrame = function(done){
  debug('.backToParentFrame()');
  this.page.switchToParentFrame();
  done();
};
```

I originally tried to make it so that once the actions inside `withFrame` are done, it would automatically go `backtoParentFrame` but I couldn't figure out how to do that, so I made it separate. So essentially you would do:

```html
<!-- For this given iframe -->
<iframe name="formIframe" src="somelogin.php" />
```

```js
nightmare
  .withFrame('formIframe', function (nightmare) {
    nightmare
      .wait('form#login')
      .type('#username', 'rclai')
      .type('#password', 'password')
      .click('#submit')
  })
  .backToParentFrame()
  .wait('.for-my-logged-in-indicator')
  .run(...);
```

molinto:
Nice feature +1

molinto:
Where would this be added @rclai please?

rclai:
You should be able to copy and paste the code I typed above and put it into `nightmare/lib/actions.js`.

molinto:
Sweet, thanks!

avioli:
+1

avioli:
Thanks @rclai I've created a plugin from your code: https://www.npmjs.com/package/nightmare-iframe

rclai:
Cool.

zeevl:
The above PR allows you to do `new Nightmare({'web-preferences': {'web-security': false}});`, which turns off same-origin policy in electron.    You can then do things like, 

```javascript
var result = yield nightmare
  .evaluate(function () {
    return $("#id_description_iframe").contents().find("body").html();
  });
```

Skit:
Thank you for the feature!
Why do I get an exception?
```console
TypeError: Cannot read property 'switchToFrame' of undefined
[...]

PR #276, opened 2015-10-10
Allow for user-specified web-preferences options.

zeevl:
In trying to turn off cross origin policy using  `web-preferences: {web-security: false}` as specified in [Electrons options](https://github.com/atom/electron/blob/master/docs/api/browser-window.md#new-browserwindowoptions), I realized it hoses nightmare's use of the `web-preferences` setting. 

This PR fixes it.   Tests included.   It also provides a v2-compatible solution for https://github.com/segmentio/nightmare/issues/203.

### DIFF
--- a/lib/runner.js
+++ b/lib/runner.js
@@ -24,19 +24,22 @@ var join = require('path').join;
 app.on('ready', function() {
 
   var win;
-  
+
   /**
    * create a browser window
    */
-  
+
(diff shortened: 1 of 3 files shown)


matthewmueller:
awesome, looks good. thanks!

nfriedly:
:+1: Can you publish a release with this?

PR #290, opened 2015-10-16
Initialize mocha-generators so tests run

chrisgeo:
Hey Nightmare Team - it looks like mocha test suite isn't automatically initializing `mocha-generators` so the tests only appear to run and succeed, instead of running.

**Branch**: Master
**Commit**: 748a2b2b14163d34e89108250e34553d991e7913

### Use-Case
Modify any test to have a `console.log` - never logs to console.

### Solution
Setup needs `require('mocha-generators').install()`

### DIFF
--- a/test/index.js
+++ b/test/index.js
@@ -2,6 +2,8 @@
  * Module dependencies.
  */
 
+require('mocha-generators').install();
+
 var Nightmare = require('..');
 var should = require('chai').should();
 var url = require('url');


LarryBattle:
:+1: It says it right there in the readme. https://github.com/vdemedes/mocha-generators

tgohn:
:+1:  need to recover test

rhysd:
This should be merged ASAP because I think currently all tests are passed wrongly.
In my environment, below tests failed after adding `require('mocha-generators').install()` as below.

```
31 passing (43s)
2 failing

1) Nightmare rendering should take a clipped screenshot:
   Error: timeout of 10000ms exceeded. Ensure the done() callback is being called in this test.


2) Nightmare rendering "after each" hook for "should take a clipped screenshot":
   Error: timeout of 10000ms exceeded. Ensure the done() callback is being called in this test.
```

matthewmueller:
yikes, just noticed this. thanks!

rhysd:
:+1:

Issue #248, opened 2015-09-25
nightmare.on('error', callback) doesn't work


antpaw:
I've added this testcase and I'm not able to pass it
https://github.com/antpaw/nightmare/commit/5e3f6617a3838eebca411db4644d04a536a5573f

antpaw:
Can you maybe give a little feedback why this is a bug and how I could fix it, or a hint where to start looking for a fix?

fritx:
https://github.com/segmentio/nightmare/blob/master/lib%2Frunner.js#L103-L105
seems it would emit `javascript` instead of `error`

antpaw:
yes, the first suggestion didn't work

edasque:
Would you use this to capture JS console errors from Electron so as to output them in node.js? How would you capture output from console.* ?

antpaw:
@edasque you mean if something is happning inside .evaluate() ? there is a `.on('javascript')` callback for js-erros and you can start your script with `DEBUG=nightmare:log` to see the logs.

btw my pr was merged, so this can be closed https://github.com/segmentio/nightmare/pull/300

edasque:
@antpaw are you sure it's DEBUG=nightmare:log ? This does nothing here. Is it just the normal DEBUG=nightmare ?

antpaw:
yes https://github.com/segmentio/nightmare/blob/master/lib/nightmare.js#L5, but there is also a '.on('log', func)' event

PR #293, opened 2015-10-18
'js-error' event added

antpaw:
Now it's possible to trace js exceptions. It solves this https://github.com/segmentio/nightmare/issues/248 issue and makes Nightmare viable for TTD again.

### DIFF
--- a/lib/preload.js
+++ b/lib/preload.js
@@ -2,6 +2,6 @@ window.__nightmare = {};
 __nightmare.ipc = require('ipc');
 __nightmare.sliced = require('sliced');
 
-window.onerror = function (errorMsg, url, lineNumber, columnNumber, error) {
-  __nightmare.ipc.send('js-error', 'Error: ' + errorMsg + ' Script: ' + url + ' Line: ' + lineNumber+':'+columnNumber, error);
-};
+window.addEventListener('error', function(errorMsg, url, lineNumber, columnNumber, error) {
+  __nightmare.ipc.send('js-error', ('Error: ' + errorMsg + ' Script: ' + url + ' Line: ' + lineNumber+':'+ columnNumber), error);
+});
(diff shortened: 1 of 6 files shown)


lib/preload.js
```diff
@@ -1,3 +1,7 @@
 window.__nightmare = {};
 __nightmare.ipc = require('ipc');
 __nightmare.sliced = require('sliced');
+
+window.onerror = function (errorMsg, url, lineNumber, columnNumber, error) {
```
matthewmueller: probably want to do `window.addEventListener('error', fn)`

otherwise the page may overwrite this.

matthewmueller:
So this would be for page errors, right? Because we already have: https://github.com/segmentio/nightmare/blob/master/lib/javascript.js#L25

antpaw:
I've changed to addEventListener. i think the keyword 'js' is fine since its used here too `.inject('js', file)`

antpaw:
I saw the callbacks for runtime errors https://github.com/segmentio/nightmare/blob/master/lib/runner.js#L106 but it's only used for `_inject` as far as i understand.
Right now i'm having a very strange issue and can not check if this two callbacks can be used as one.
The problem is, all the tests are passing within 400ms even if i have something like this `  title.should.equal('NOPE')` and i have 30 instances of Electron open in my doc. Even if i freshly checkout master form your remote, mocha tests are just behaving crazy. Do you have any idea why this could happen?

antpaw:
Oh it looks like this issue is happing everywhere 
`$ make test` in this old build took 29s https://circleci.com/gh/segmentio/nightmare/364
`$ make test` in this build took 1s https://circleci.com/gh/segmentio/nightmare/373
since build 365 tests don't really run
what could have caused this? the commit for build 365 https://github.com/segmentio/nightmare/commit/58ff6acd00838576c50f953473101ec9241cb60f has nothing at all

antpaw:
ok i've fixed this problem here https://github.com/segmentio/nightmare/pull/299

antpaw:
Sorry i need to move this pr to a new branch https://github.com/segmentio/nightmare/pull/300

mrself:
Looks like this does not work now. I manually created a page with a non-valid script and I can not track it with nitghtmare.

PR #296, opened 2015-10-19
Allow the user to disable Chromium's HTTP cache

peterjanes:
This introduces a new `browser-configure` event that happens before `browser-initialize`, which allows the user to configure the Chromium instance itself. Might be useful to provide other flags, but I only needed `disable-http-cache`. (Apologies for the lack of tests.)

### DIFF
--- a/lib/nightmare.js
+++ b/lib/nightmare.js
@@ -67,6 +67,7 @@ function Nightmare(options) {
 
   this.child = child(this.proc);
   this.child.once('ready', function() {
+    self.child.emit('browser-configure', options);
     self.child.once('browser-initialize', function() {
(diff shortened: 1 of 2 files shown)


kevinprotoss:
+1

inversion:
+1

matthewmueller:
sweet, we should probably namespace this like we did with paths. `{ chromium: { http-cache: false } }` maybe? 

anyone have any other ideas?

inversion:
Close to reduce entropy? You can now pass switches to the runner.

matthewmueller:
yep, this can be solved using switches. thanks for the PR!

PR #298, opened 2015-10-20
Removed Makefile and added test to package.json

LarryBattle:
- Removed `Makefile`
- `npm test` can now be used to run tests
- Updated Readme.md to mention how to run tests
- tmp folders are now created inside `test` instead of `/tmp/nightmare`
- Includes https://github.com/segmentio/nightmare/pull/290


matthewmueller:
personally like having the makefile for `make test`, but it's kind of frivolous for this case anyway. thoughts @reinpk ?

tejasmanohar:
@matthewmueller i'd recommend #303 instead. you may have more make tasks later but `npm test` is very common (as the shortcut is even there from `npm run-script` / `npm run` test)... so best to support that. i think many CI systems default to this, too.

matthewmueller:
@tejasmanohar yep, i like that more. I think we should keep the makefile, but the other stuff looks good!

tejasmanohar:
yep! makefile is good, we can do much more there. `npm test` is just a helper :)

:-1: on this one

matthewmueller:
thanks @LarryBattle this is on master now.

PR #300, opened 2015-10-20
Excetions and console.logs now have event callbacks

antpaw:
There are two ways of getting an exception in Nightmare, one is to inject some "bad" code via .evaluate(), this will stop nightmare and even call done() in mocha. But there is also the issue with already existing code, that has errors, now there is an option to listen for this.
Same with console log, before only console logs from evaluate() where shown.

### DIFF
--- a/lib/preload.js
+++ b/lib/preload.js
@@ -2,6 +2,6 @@ window.__nightmare = {};
 __nightmare.ipc = require('ipc');
 __nightmare.sliced = require('sliced');
 
-window.onerror = function (errorMsg, url, lineNumber, columnNumber, error) {
-  __nightmare.ipc.send('js-error', 'Error: ' + errorMsg + ' Script: ' + url + ' Line: ' + lineNumber+':'+columnNumber, error);
-};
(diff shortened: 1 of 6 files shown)


lib/preload.js
```diff
@@ -1,3 +1,13 @@
 window.__nightmare = {};
 __nightmare.ipc = require('ipc');
 __nightmare.sliced = require('sliced');
+
+window.addEventListener('error', function(e) {
+  __nightmare.ipc.send('js-error', e.message, e.error.stack);
+});
+
+defautlLog = console.log;
```
matthewmueller: `defautl => default`

matthewmueller:
Sweet! So I thought about this a bit more. One issue I have with the event name:

1) it's inconsistent with the existing event names: https://github.com/atom/electron/blob/master/docs/api/browser-window.md#events

I think for these this reason, it should be something like: `page-error` and `page-log`.

The only other thing that would be good is squashing the commit, so it's more concise in the commit log. 

Otherwise looks good, thanks!

antpaw:
ok, done

matthewmueller:
Thanks @antpaw !

PR #301, opened 2015-10-20
prefer fn#bind(this) to var self = this

tejasmanohar:
i'd say that since everyone using nightmare is es5+, Function#bind should be preferred to the `var self = this` strategy since rebinding `this` is the end goal and we don't have to declare another variable.

thoughts?

### DIFF
--- a/lib/actions.js
+++ b/lib/actions.js
@@ -304,10 +304,9 @@ exports.inject = function (type, file, done) {
     var js = fs.readFileSync(file, { encoding: 'utf-8' });
     // TODO: fix double-injection hack that magically makes
     // jQuery injection succeed :(
-    var self = this;
-    self._inject(js, function () {
-      self._inject(js, done);
-    });
+    this._inject(js, function () {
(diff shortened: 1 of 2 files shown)


matthewmueller:
whenever i see this change, i always wonder about the benchmarks, since we're wrapping every function in another function, as opposed to just points to the `this` ref. do you have any benchmarks of this being a negligible change? doesn't have to be nightmare specific but just in general?

rosshinkley:
@matthewmueller FWIW, there's a decent looking [jsperf](https://jsperf.com/bind-vs-closure-setup/49).  According to that, it looks like the `bind` method is significantly slower.  Is that close to what you had in mind?

matthewmueller:
@rosshinkley yep this is exactly what i was looking for:

img

even though it's a micro optimization, i don't think using `self` is much worse syntactically.

tejasmanohar:
aha, nope- i did not have a stat for this but am glad to see this one- didn't know...

PR #302, opened 2015-10-20
prefer module.exports to exports for consistency + convention + clarity

tejasmanohar:


LarryBattle:
`exports` and `module.exports` point to the same object unless changed. 
- http://stackoverflow.com/questions/7137397/module-exports-vs-exports-in-node-js

It would be better to change the 2 references to `module.exports` to `exports`.

tejasmanohar:
It would be better to stick to `module.exports` because that's the convention. `exports` is merely a shorthand that can sometimes cause confusion.

For example, if you use `module.exports` and `exports`, what's exported? And when you do, `module.exports = exports = x`, that's just additional meaningless clutter to your codebase especially when `module.exports` *always* works in Node.js :wink:

matthewmueller:
leaning -1 on this change, though fairly indifferent. the reason for this is because it's added syntax and i think most people in the node community (also webpack and browserify world) understand commonjs at this point. 

i do agree that `exports` shouldn't have made it into the spec though.

tejasmanohar:
alright. regardless of convention, i'd still recommend against `exports`... but i understand commonjs so whatever. i myself am +- 0

matthewmueller:
closing this. i think we're going to stick wtih `exports` for the time being. thanks!

tejasmanohar:
fair

PR #305, opened 2015-10-21
.wait(fn, arg1, arg2, ...) accepts arguments now

antpaw:
... just like .evaluate()


chrisgeo:
:+1: I was going to carve out time to write this as well.

a7madgamal:
we need that :pray:

Issue #201, opened 2015-07-31
add eventType to click() for 'mousedown' support


abe-winter:
I'm working with a page that needs `'mousedown'`, not `'click'`events. Any interest in adding an `eventType` param to Nightmare.click? (would default to eventType='click').

Change would be [here](https://github.com/segmentio/nightmare/blob/073fa6dc02b84ab6762805d4b40239aa9fa7d57b/lib/actions.js#L193)

Not sure how this overlaps with the new `mouseover` action.

rosshinkley:
Fixed in #308.  Closing.

PR #308, opened 2015-10-21
enhancement: added mousedown event

nicholasblexrud:
Added a ```mousedown``` event to fulfill enhancement #201.


matthewmueller:
sorry for the late response. this is on master now. thanks!

Issue #278, opened 2015-10-11
blocking function


jney:
doing the following code, `console.log('after')` is called 1/10 times. most of the time it stays blocked

```es6
  const nightmare = Nightmare()
  console.log('before')
  const objects = yield nightmare
    .goto(url)
    .evaluate(function () {
      return 'noop'
    })
  console.log('after')

  yield nightmare.end()
  return objects
```

matthewmueller:
what's the url?

jney:
`http://fd5-courses.leclercdrive.fr/magasin-069401-Vitry-sur-Seine.aspx`

rosshinkley:
@jney Is this still an issue?

jney:
i just tried it now (i don't use nightmare for now) and it works !
thank you

Issue #251, opened 2015-09-25
wait(fn) stops evaluating after a few cycles


FdezRomero:
Hello there,

I have been trying to use `.wait(fn)` and my script always got stuck at this point, so I started debugging the `waitfn()` function in `lib/actions.js`, adding some `debug()` like this:

```javascript
function waitfn (self, fn, done) {
  debug('waitfn');
  self._evaluate(fn, function (err, result) {
    debug('_evaluate');
    if (result) {
      debug('Wait is done');
      return done();
    }
    else {
      debug('Waiting another 250ms');
      setTimeout(function () {
        waitfn(self, fn, done);
      }, 250);
    }
  });
}
```

And noticed that `waitfn()` stops iterating after a few cycles. This is the output:

```
  nightmare:actions .wait() for fn +26ms
  nightmare:actions waitfn +0ms
  nightmare:actions _evaluate +3ms
  nightmare:actions Waiting another 250ms +1ms
  nightmare:actions waitfn +254ms
  nightmare:actions _evaluate +2ms
  nightmare:actions Waiting another 250ms +1ms
  nightmare:actions waitfn +251ms
  nightmare:actions _evaluate +191ms
  nightmare:actions Waiting another 250ms +0ms
  nightmare:actions waitfn +252ms
  // missing a call to "_evaluate" here
```

Any idea why this is happening? `_evaluate` only gets called 2 or 3 times and I really don't know why...

Thanks,
Rodrigo

tjokimie:
I'm also facing the exact same problem. `waitfn` stop iterating after 1-3 iterations. If I increase the timeout from `250` to 2500 or more the script executes successfully.

CatTail:
Any progress on this issues?

kazinov:
+1

rosshinkley:
@FdezRomero @CatTail @kazinov @tjokimie I cannot reproduce this problem.  Has this issue been resolved?

rosshinkley:
I'm going to consider this issue resolved.  If this is still a problem, feel free to reopen/submit a new issue.

deffwe:
seems that the bug isn't resolved yet, i'm on version 2.10. anybody has solution? thx

rosshinkley:
@deffwe If you have a minimum repro, post it and I'll reopen this issue to take a look.

Issue #238, opened 2015-09-21
Using nightmare for Load/Stress testing


ndamjan:
Hello, I am trying to use nightmare to perform load and stress testing of a web application where e.g. 20 concurrent users are expected. However, it appears that only 1 session is active and others are simply waiting, although I can see 20 phantomjs processes in task manager (using Windows 7).
Did anyone else also try this, does nightmare support such use case at all? Thanks for feedback!

Here is my script:

``` javascript
var Nightmare = require('nightmare');

var createTest = function (id) {
    return new Nightmare({timeout: 7200000})
    .log('#' + id + ' << BEGIN >>')
    .log('#' + id + ' ' + 'Starting to load')
    .goto('http://192.168.1.1:8080/webapp/index.jsf')
    .log('#' + id + ' ' + 'Page loaded')
    .log('#' + id + ' ' + 'Enter user/pass')
    .type('input[id="loginForm:tab:username"]', 'test')
    .type('input[id="loginForm:tab:password"]', 'xxxxxx')
    .screenshot('test' + id + ' ' + 'screen-1.png')
    .log('#' + id + ' ' + 'Click Login button')
    .click('button[name="loginForm:tab:btnLogin"]')
    .wait()
    .screenshot('test' + id + ' ' + 'screen-2.png')
    .log('#' + id + ' ' + 'Click APP button') 
    .click('button[name="overviewForm:j_id_d:j_id_q"]')
    .wait()
    .log('#' + id + ' ' + 'Enter 123456')
    .type('input[id="Search:j_id_c:j_id_s"]', '123456')
    .log('#' + id + ' ' + 'Click Search button') 
    .click('button[name="Search:j_id_c:j_id_z"]')
    .screenshot('test' + id + ' ' + 'screen-3.png')
    .wait('form[name="App:SomeInfo"]')
    .wait(2000) // wait 2 more seconds to finish the chart animation
    .screenshot('test' + id + ' ' + 'screen-4.png')
    .log('#' + id + ' ' + 'View search page')
    .log('#' + id + ' << END >>');
}

var runTest = function (test) {
    test.run(function (err, nightmare) {
      if (err) return console.log(err);
    });
}

var tests={};

var numTests = 20;

console.log('Preparing ' + numTests + ' tests');

for (i=1; i<=numTests; i++) {
    tests[i] = createTest(i);
}

console.log("Executing tests");

for (i=1; i<=numTests; i++) {
    runTest(tests[i]);
}
```

[...]

PR #320, opened 2015-10-27
Use "did-stop-loading" event in "continue"

fr-:
The formerly used "did-finish-load" is not always emitted after an action
in case webContents.isLoading() is true. Thus the "continue" event is never
emitted. The probability of this behavior rises with rising frequency
of actions, resulting in flaky behavior and hangs depending on the
target site.

The electron documentation states that "did-finish-load" is only
dispatched when the "onload" event is dispatched and the spinner of the
tab has stopped spinning. Yet the "onload" event is not always dispatched
when webContents.isLoading() is true. Therfore the correct event is
"did-stop-loading" which is always dispatched after isLoading()
transitions to false.

### DIFF
--- a/lib/runner.js
+++ b/lib/runner.js
@@ -183,7 +183,7 @@ app.on('ready', function() {
       ready();
     } else {
       parent.emit('log', 'navigating...');
-      win.webContents.once('did-finish-load', function() {
+      win.webContents.once('did-stop-loading', function() {
         parent.emit('log', 'navigated to: ' + win.webContents.getUrl());
         ready();
       });


fr-:
As far as I can see this will fix a few open issues for example: #278 #238 #251

matthewmueller:
Wow. This would explain a lot of the inconsistencies.

I *think* we should be good on the onload stuff since, we're injecting JS at the bottom of the page, but can you think of any case where not waiting for the `onload` event could break things?

/cc @reinpk

fr-:
There might be timing issues that could break _some_ existing code that relies on the current behavior of ``continue`` only emitting on ``did-finish-load`` and never earlier, but you could probably formulate that code differently using direct event listeners on ``did-finish-load``. Furthermore such code does seem to be incorrect to rely on that logic in the first place, so it should probably be changed in any case (or the interface should expose the intent directly).

Apart from that I am pretty sure that even though the ``did-finish-load`` event always precedes ``did-stop-loading`` it also _always_ coincides with the latter, which in turn gives us compatibility with the original (intended?) case.

Of course, there could be edge cases that I am not aware of right now. If I find the time I can look into the electron code to graph out all event sequences possible, but if anyone knows right off the bat, that would be awesome.

reinpk:
i'm good with this!

kazinov:
Hi. This solves my issues with wait functions. Are you going to merge it?

inversion:
Seems to work well, would be nice to see it merged.

reinpk:
yeah sorry haven't had time to read over the code yet, soon sorry!

Namek:
@reinpk it's just one line!

jperl:
Applying this change manually fixed #322 for me

rosshinkley:
+1.  Wait is much more reliable with this patch.

matthewmueller:
Okay going to merge this as it seems like it's helping out a lot of folks. Haven't dug into the internals yet, but I'm sure we can make changes if issues crop up.

Great work @fr- !

PR #336, opened 2015-11-09
Nightmare.action

### LINKED ISSUE #167: Nightmare.action(name, method)
Might be good for extensibility. So things like:

```js
Nightmare.action('cookies', function(callback, done){
  debug('.cookie()');
  var self = this;
  this.page.get('cookies', function(cookies){
    var newCookies = callback(cookies);
    if(newCookies){
      self.page.set('cookies', newCookies, done);
    }else{
      done();
    }
  });
});
```

Could be done without necessarily adding it to core. Basically just the inners of this:

```js
Object.keys(actions).forEach(function (name) {
  var fn = actions[name];
  Nightmare.prototype[name] = function() {
    debug('queueing action "' + name + '"');
    var args = [].slice.call(arguments);
    this.queue.push([fn, args]);
    return this;
  };
});
```

inversion:
fixes #167

### DIFF
--- a/lib/nightmare.js
+++ b/lib/nightmare.js
@@ -255,16 +255,23 @@ Nightmare.prototype.then = function(fulfill, reject) {
   return this;
 };
 
-/**
- * Attach all the actions.
- */
 
-Object.keys(actions).forEach(function (name) {
-  var fn = actions[name];
+Nightmare.action = function(name, fn) {
+
   Nightmare.prototype[name] = function() {
     debug('queueing action "' + name + '"');
     var args = [].slice.call(arguments);
     this._queue.push([fn, args]);
     return this;
   };
+};
+
+/**
+ * Attach all the actions.
+ */
+
+Object.keys(actions).forEach(function (name) {
+  var fn = actions[name];
+
+  Nightmare.action(name, fn);
 });


Namek:
<3 please merge it.

chrisgeo:
Probably needs a test case before merging.

inversion:
It's conceptually and implementation-wise very simple so I'm not sure about testing being needed. One caveat that could be added to the docs is that if you have actions which themselves call other actions you'll need to resolve them, otherwise you end up just queueing an action which itself queues actions. I've generally been handling this by passing a `done` callback to my actions and calling `then` at the end of my action code.

matthewmueller:
Haha when I opened that issue, I forgot about `.use(...)`. What do you guys like better `Nightmare.action(action, event)` or `Nightmare.use(fn)`?

Going to rope in the .use people too.

inversion:
I've never 'used' `use` but I think there's a case for breaking out some actions into plugins. For example I have got a growing suite of mobile-specific actions that might end up bloating nightmare if merged here.

I also think we should handle conflicts on the prototype in either case and namespacing to plugins seems like a good way to do that.

matthewmueller:
just added both `use` and `action`. Here's some more information: #376 376

matthewmueller:
Thanks for the PR! I incorporated these changes in #376

PR #337, opened 2015-11-09
console.warn and console.error events

inversion:
I've named the events `page-console-error` and `page-console-warn` because otherwise the error event has a naming conflict with the existing event.

I've suggested deprecating page-log and making it page-console-log to bring it inline with the other new events. Any thoughts?

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -154,13 +154,18 @@ Returns whether the selector exists or not on the page.
 Returns whether the selector is visible or not
 
 #### .on(event, callback)
-Capture page events with the callback. You have to call `.on()` before calling `.goto()`. Supported events are [documented here](http://electron.atom.io/docs/v0.30.0/api/browser-window/#class-webcontents). Additional to the electron-events we provide nightmare-events `'page-error'` and `'page-log'`.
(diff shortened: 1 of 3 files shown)


matthewmueller:
Hm, can you explain the use case a little better? I was sort of thinking about it in terms of stdout and stderr. I believe console.warn goes to stdout. 

The downside I can see with this is that you might end up missing events if you don't hook into the correct events. Also what about for events coming from `console.info` and `console.debug`.

inversion:
Hi @matthewmueller , the use case for us is that our application logs messages to error, warn, info, log depending on severity and these are useful to capture in our tests (we're mainly using nightmare for UI testing). I think a nicer implementation would be a generic `console` event which captures all console output but also adds a `type` or similar field indicating the 'level' (dir, trace, info, error) etc.

Happy to implement that if it sounds like something that would be merged.

matthewmueller:
yep I love it: `nightmare.on('console', function (level, msg, ...) { ... })`

inversion:
Done in https://github.com/segmentio/nightmare/commit/f499ddfac65416a781d5c17c86e4fee33719bcab

PR #339, opened 2015-11-10
add method for set position of window.

lexifdev:
Add method for set position of window.

### DIFF
--- a/lib/actions.js
+++ b/lib/actions.js
@@ -335,6 +335,20 @@ exports.viewport = function (width, height, done) {
 };
 
 /**
+ * Set the position.
+ *
+ * @param {Number} x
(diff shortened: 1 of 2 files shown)


matthewmueller:
may i ask what the usecase is here? the window should be hidden in nightmare, so i'm not sure why you'd want to move it.

lexifdev:
recently I use nightmarejs with robotjs for complex test that needed real mouse event.
robotjs is uses absolute position for move mouse cursor, so test is need reset the window position before start.

rosshinkley:
@lexifdev With the additions to `.action()` with #425, this could be implemented as a plugin.  Feel free to open up an issue or join the gitter room if you have questions.

Closing.

PR #351, opened 2015-11-18
point to versioned Electron documentation

rosshinkley:
The documentation in Electron's master branch differs from the documentation for the version of Electron Nightmare is on.  As such, the link in the Nightmare constructor should point to the versioned documentation.


matthewmueller:
awesome, thanks!

PR #353, opened 2015-11-24
Return a buffered screenshot if no path is provided

atonparker:
I've got a use case for the screenshot function that doesn't require it to be written to disk. Electron itself returns the image in memory, so I thought it'd be useful if writing it to the disk were optional. This pull request maintains backwards compatibility and adds the necessary tests. These tests actually pass (https://github.com/segmentio/nightmare/pull/290), but I didn't include the line from that PR to avoid a conflict.

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -162,8 +162,8 @@ This event is triggered if any javscript exception is thrown on the page. But th
 ##### .on('page-log', errorMessage, errorStack)
 This event is triggered if `console.log` is used on the page. But this event is not triggered if the injected javascript code (e.g. via `.evaluate()`) is using `console.log`.
 
-#### .screenshot(path[, clip])
-Saves a screenshot of the current page to the specified `path`. Useful for debugging. The output is always a `png`. You can optionally provide a clip rect as [documented here](https://github.com/atom/electron/blob/master/docs/api/browser-window.md#wincapturepagerect-callback).
+#### .screenshot([path, clip])
+Takes a screenshot of the current page. If a path is provided, it saves it to the specified `path`. Otherwise it returns a `Buffer` of the image data. Useful for debugging. The output is always a `png`. You can optionally provide a clip rect as [documented here](https://github.com/atom/electron/blob/master/docs/api/browser-window.md#wincapturepagerect-callback).
 
 #### .pdf(path, options)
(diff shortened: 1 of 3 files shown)


Readme.md
```diff
@@ -162,8 +162,8 @@ This event is triggered if any javscript exception is thrown on the page. But th
 ##### .on('page-log', errorMessage, errorStack)
 This event is triggered if `console.log` is used on the page. But this event is not triggered if the injected javascript code (e.g. via `.evaluate()`) is using `console.log`.
 
-#### .screenshot(path[, clip])
-Saves a screenshot of the current page to the specified `path`. Useful for debugging. The output is always a `png`. You can optionally provide a clip rect as [documented here](https://github.com/atom/electron/blob/master/docs/api/browser-window.md#wincapturepagerect-callback).
+#### .screenshot([path, clip])
```
matthewmueller: Hmm, I think the way this signature is documented is a little confusing. Would this make more sense?

```
.screenshot([path|clip], [clip])
```

matthewmueller:
Awesome! Just want to figure out how we should document this change.

Readme.md
atonparker: [Here's](https://nodejs.org/api/fs.html#fs_fs_watch_filename_options_listener) some inspiration from the node API. They use

    fs.watch(filename[, options][, listener])

for a function that can take `filename`, or `filename, options`, or `filename, listener`, or all three. So how about

    .screenshot([path][, clip])

I can also add a two or three line example of how the path, the clip, or both can be omitted.

atonparker:
I updated the documentation to match the official node.js docs for a function with two optional parameters. I also tried to add some clarifying language. Thoughts?

lib/actions.js
```diff
@@ -371,18 +371,22 @@ exports.scrollTo = function (y, x, done) {
  * @param {Object} clip
  * @param {Function} done
  */
-
+``
```
matthewmueller: woops, look like a typo

matthewmueller:
sorry for the late response on this, looks good. just need to fix up the typo and we're good to go

atonparker:
Hah yep, that's a typo. Should be good to go now.

matthewmueller:
thanks @atonparker !

PR #355, opened 2015-11-25
Add cookies getter function

vially:
This is useful for retrieving HttpOnly cookies which are not available from `document.cookie`.


matthewmueller:
thanks! this has been added via `nightmare.cookies.get(name)` in #360

Issue #328, opened 2015-11-02
Complete screenshot (no scroll)


chipay:
I've been playing around with the screenshot functionality but I couldn't make it print the whole content.
My use case is to compare the whole site snapshot for differences (similar to the phantomcss library). Casperjs allows to capture the whole scrollable screen in a png. Is there a way of doing the same with nightmare? If not, I'll see if it can be added easily but it may take some time as I just started getting into nightmare.

DWboutin:
^ +1

audable:
+1

rosshinkley:
You can ask Nightmare to get the dimensions of the body element (using `scrollHeight`/`scrollWidth`), and then use `Nightmare.viewport` to set the viewport dimensions.  After that, you should be able to take a screenshot of the entire page.  I'm not sure this will work 100% of the time, but it's what I've got working for me and I thought I'd share.

An example might be useful:

```javascript
var Nightmare = require('nightmare'),
    vo = require('vo');

function * run() {
    var nightmare = new Nightmare({
        show: false,
        width: 1024,
        height: 768
    });
    var dimensions = yield nightmare.goto('http://www.gmail.com')
        .wait('body')
        .evaluate(function() {
            var body = document.querySelector('body');
            return { 
                height: body.scrollHeight,
                width:body.scrollWidth
            }
        });

    console.dir(dimensions);
    yield nightmare.viewport(dimensions.width, dimensions.height)
        .wait(1000)
        .screenshot(require('path')
            .join(__dirname, 'gmail.png'));

    yield nightmare.end();
}   

vo(run)(function() {
    console.log('done');
});
```

A couple of notes and questions:

1. The width (and to a lesser extent, height) of the original Nightmare instance should be reasonably big, especially if your site is responsive.  Make sure the width is near the threshold you're targeting your screenshot for.
1. `show` has to be false.  If Electron actually renders to a real screen and the page scrolls off the bottom, the viewport will render with the scrollbar as it appears on screen.  I _think_ this is because `show:false` renders in the framebuffer which has no screen dimensions to honor, but I'm not sure.  Anyone know for certain?
[...]

PR #358, opened 2015-11-30
screenshotSelector  function for taking a screenshots by element selector

Kreozot:
screenshotSelector function (with test and readme)

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -165,6 +165,9 @@ This event is triggered if `console.log` is used on the page. But this event is 
 #### .screenshot(path[, clip])
 Saves a screenshot of the current page to the specified `path`. Useful for debugging. The output is always a `png`. You can optionally provide a clip rect as [documented here](https://github.com/atom/electron/blob/master/docs/api/browser-window.md#wincapturepagerect-callback).
 
+#### .screenshotSelector(path, selector)
+Saves a screenshot of the element by the selector to the specified `path`. The output is always a `png`.
+
 #### .pdf(path, options)
 Saves a PDF with A4 size pages of the current page to the specified `path`. Options are [here](http://electron.atom.io/docs/v0.30.0/api/browser-window/#webcontents-printtopdf-options-callback).
 
--- a/Readme.md
+++ b/Readme.md
@@ -201,8 +201,8 @@ This event is triggered if `alert` is used on the page.
 #### .screenshot([path][, clip])
 Takes a screenshot of the current page. Useful for debugging. The output is always a `png`. Both arguments are optional. If `path` is provided, it saves the image to the disk. Otherwise it returns a `Buffer` of the image data. If `clip` is provided (as [documented here](https://github.com/atom/electron/blob/master/docs/api/browser-window.md#wincapturepagerect-callback)), the image will be clipped to the rectangle.
 
-#### .screenshotSelector(path, selector)
-Saves a screenshot of the element by the selector to the specified `path`. The output is always a `png`.
+#### .screenshotSelector([path,] selector)
+Takes a screenshot of the element by the selector. The output is always a `png`. If `path` is provided, it saves the image to the disk. Otherwise it returns a `Buffer` of the image data.
 
 #### .pdf(path, options)
 Saves a PDF with A4 size pages of the current page to the specified `path`. Options are [here](http://electron.atom.io/docs/v0.30.0/api/browser-window/#webcontents-printtopdf-options-callback).
--- a/lib/actions.js
+++ b/lib/actions.js
@@ -423,7 +423,7 @@ exports.screenshot = function (path, clip, done) {
 };
 
 /**
- * Take a screenshot of element by selector.
+ * Take a screenshot of an element by selector.
  *
  * @param {String} path
  * @param {String} selector
@@ -432,6 +432,11 @@ exports.screenshot = function (path, clip, done) {
 
 exports.screenshotSelector = function (path, selector, done) {
   debug('.screenshotSelector()');
+  if (typeof selector === 'function') {
+    done = selector;
+    selector = path;
+    path = undefined;
+  };
   var self = this;
 
   this._evaluate(function (selector) {
@@ -452,7 +457,7 @@ exports.screenshotSelector = function (path, selector, done) {
     self.child.once('screenshot', function (img) {
       var buf = new Buffer(img.data);
       debug('.screenshotSelector() captured with length %s', buf.length);
-      fs.writeFile(path, buf, done);
+      path ? fs.writeFile(path, buf, done) : done(null, buf);
     });
     self.child.emit('screenshot', path, clip);
   }, selector);
--- a/lib/actions.js
+++ b/lib/actions.js
@@ -388,6 +388,42 @@ exports.screenshot = function (path, clip, done) {
 };
 
 /**
+ * Take a screenshot of element by selector.
+ *
+ * @param {String} selector
+ * @param {Function} done
+ */
+
+exports.screenshotSelector = function (path, selector, done) {
+  debug('.screenshotSelector()');
+  var self = this;
+
+  this._evaluate(function (selector) {
+    var element = document.querySelector(selector);
+    if (element) {
+      var rect = element.getBoundingClientRect();
+      return {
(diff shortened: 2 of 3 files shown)


yhuard:
I was working on the same feature and was about to push it when I saw your PR. This is a much needed feature, I hope it will be merged soon. Thank you @Kreozot for your work!

FYI here's my own implementation: 

```js
exports.captureSelection = function (path, selector, done) {
  debug('.captureSelection()');
  var self = this;
  this.evaluate(function (selector, clip) {
    var element = document.querySelector(selector);
    if (element) {
      var rect = element.getBoundingClientRect();
      return {
        x: rect.left,
        y: rect.top,
        width: rect.width,
        height: rect.height
      };
    }
    return false;
  }, selector)
    .then(function (clip) {
      if (clip) {
        self.child.once('screenshot', function (img) {
          var buf = new Buffer(img.data);
          debug('.captureSelection() captured with length %s', buf.length);
          fs.writeFile(path, buf, done);
        });
        self.child.emit('screenshot', path, clip);
      }
    }, function (err) {
      console.log(err);
    });
};
```

But your solution seems at least as good

Kreozot:
Thanks. In my implementation I also used Math.round function because clip parameters must be integer but getBoundingClientRect can return decimal values.

yhuard:
Nice catch, I was having an issue making Electron crash and adding Math.round solved it

yhuard:
It would be nice to add an optional parameter which would give more control to the developer on the way screenshot are taken. What I'm thinking about is:

```javascript
{
  offsetX: horizontal distance from the top left corner of the selection,
  offsetY: vertical distance from the top left corner,
  width: force the width instead of using selection's width
  height: force the height
}
```

If you can think of any other options, I'd be glad to talk about them with you

Kreozot:
I think this is redundant. You can not guess all use cases. I think all individual use cases must be implemented outside this function.
First time I've wanted to add scroll functionality to this function, but then I've decided to keep it clean ([YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it) principle)

yhuard:
Scroll functionality is another thing, it should be done outside of this function, I totally agree with this. In fact I had a use case for those options (taking the same screenshot at different times to catch visual regressions) but I can go without them as well, it's not a problem for me

sepehrs:
I think scroll is needed for this function. Otherwise you could call the function and get a screenshot of the wrong section of the page.

I don't think the user should have to think about manually scrolling before calling this function.

If you want to keep it clean maybe create a separate function such as scrollToElement which is called internally by captureSelection.

For extra points you could get it to scroll back to the pervious position after the screenshot was taken.

Something like this inside of eval should do it.

```javascript
window.scrollTo(Math.round(rect.left), Math.round(rect.top));
```

You just need to remember that scrolling is not instantaneous and you might need to wait a little bit before taking the screenshot.

chipay:
I find it hard to think of a use case where it's necessary to take a snapshot of a selector without scrolling there. I think most use cases that require a snapshot of a selector is because they want an image of the content of that selector. You can't get that without scrolling first. 
What do you think guys? I think that's a more sensible use case.

yhuard:
I still think that it should be done manually before calling `screenshotSelector`. Maybe it would be interesting to implement a similar `scrollToSelector` function, because of two things:
- a function should focus on doing only one thing and doing it well ;
- what if the user doesn't want to scroll to the element before capturing the screenshot? there might be use cases for that

Kreozot:
@yhuard :+1: 
scrollToSelector PR is coming

sepehrs:
If you call screenshotSelector('header') and you have scrolled to the bottom of the page you will get a screenshot but it will be of the wrong thing.

I can not think of a use case where you say give me a screenshot of an element but bring me back a screenshot thats the same as that element but not really of that element.

I agree with concept of a function having to do one thing and doing it well.

I would then suggest having the following functions.

1. getSelectorRect
2. scrollTo (which already exists) / scrollToSelector
3. screenshot (with rect as a parameter which exists already)

then a wrapper for these 3 functions called screenshotSelector.

Then if you want an additional function for scrolling to a selector you can call the getSelectorRect and use it in conjunction with scrollTo.

But it still really confusing if you call screenshotElement and getting a random screenshot back.

rosshinkley:
+1 for @sepehrs' suggestion: `screenshotSelector` should not violate the principle of least surprise.  If you ask for a picture of a specific element, you should get a shot of the whole element.  I completely agree with splitting the underlying functionality into their own functions, then having `screenshotSelector` be a wrapper for those functions

That being said, there are corner cases that I have mixed feelings about.  For example, taking a complete screenshot when the body is longer or wider than the viewport rectangle (as with #328) - should `screenshotSelector` take this into account and resize the viewport automatically, putting it back upon screenshot completion?  Thoughts?

yhuard:
@rosshinkley I've already thought of that and ended up thinking that resizing the viewport is dangerous: it can have side effects that the user didn't expect. For example, if an element is too big for, let's say, a medium screen but is hidden on large screen and we apply the resizing, the element may disappear. 

We define the viewport at beginning because we want to test the rendering with that viewport. If the screenshot cuts the element we capture, it's actually a good indicator that something's wrong with our style!

sepehrs:
I wasn't actually talking about reszing the viewport. In my opinion there is no need for that since you would want to see how the element would look on the device (viewport) you have selected.

I still don't see any problems with scrolling to the element.

Kreozot:
Just created PR with scrollToSelector function: #362

Kreozot:
Just created PR with scrollToSelector function: #363

rosshinkley:
@yhuard @sepehrs Like I said, I had mixed (and admittedly, mostly negative) feelings about resizing the viewport.  I agree that it's not a great idea, but mileage may vary with elements that extend beyond the viewport.

sepehrs:
@rosshinkley the reason i didn't mind to much about the reszing of the viewport is the point @yhuard made about css selector but also a genuine use case could be that you want to see how the element looks on a specific device. if for example you are taking a screenshot of a header on mobile. you can see exactly what the above the fold content is.

Kreozot:
@matthewmueller, I think you can merge this now if you don't have any suggestions.
After making "path" parameter optional it became impossible to overload original "screenshot" function by checking parameter type as I mentioned in our discussion in #363. So I think this PR is ready to be merged.

matthewmueller:
the more i think about this, the more i think this makes sense as a plugin. i'll be adding plugin support via `.use(...)` today

felixcrive:
Oh, That's good news for me! Thanks @matthewmueller !

matthewmueller:
Okay, I've added `Nightmare.action` to support this PR. Here's how you'd do it:

```js
Nightmare.action('screenshotSelector', function (path, selector, done) {
  debug('.screenshotSelector()');
  if (typeof selector === 'function') {
    done = selector;
    selector = path;
    path = undefined;
  };
  var self = this;

  this._evaluate(function (selector) {
    var element = document.querySelector(selector);
    if (element) {
      var rect = element.getBoundingClientRect();
      return {
        x: Math.round(rect.left),
        y: Math.round(rect.top),
        width: Math.round(rect.width),
        height: Math.round(rect.height)
      };
    }
  }, function (a, clip) {
    if (!clip) {
      throw new Error('invalid selector "' + selector + '"');
    }
    self.child.once('screenshot', function (img) {
      var buf = new Buffer(img.data);
      debug('.screenshotSelector() captured with length %s', buf.length);
      path ? fs.writeFile(path, buf, done) : done(null, buf);
    });
    self.child.emit('screenshot', path, clip);
  }, selector);
})

// run
yield Nightmare()
  .goto('http://google.com')
  .screenshotSelector('google.png', 'title')
```

Does this work for you?

felixcrive:
@matthewmueller I will give it a try this morning and I will let you know. The selector I want to capture is an external iframe. This is great. Thanks!

felixcrive:
@matthewmueller It did work, thanks a lot!

felixcrive:
Is there a way to capture an image if the evaluation of the page returns true?

matthewmueller:
@felixcrive, sorry i'm not sure what you mean, do you mind showing me what you mean?

reinpk:
best supported as a plugin

PR #361, opened 2015-12-02
Add sendInputEvent, update to latest Electron

inversion:
sendInputEvent wasn't implemented in Electron 0.33 so I bumped the dependency, tests seem ok. Happy to add a test for sendInputEvent if needed.

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -111,6 +111,9 @@ Clicks the `selector` element once.
 #### .type(selector, text)
 Enters the `text` provided into the `selector` element.
 
+#### .sendInputEvent(event)
+Send the input event to the page. See [Electron Docs](https://github.com/atom/electron/blob/master/docs/api/web-contents.md#webcontentssendinputeventevent).
+
 #### .check(selector)
 Toggles the `selector` checkbox element.
 
--- a/lib/actions.js
+++ b/lib/actions.js
@@ -406,3 +406,17 @@ exports.pdf = function (path, options, done) {
   });
   this.child.emit('pdf', path, options);
 };
+
+/**
+ * Send an inputEvent.
+ *
+ * @param {Object} event
+ */
+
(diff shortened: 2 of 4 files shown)


matthewmueller:
Awesome! So we're planning on making this a little more high-level, so `.click(...)`, `.rightclick(...)`, etc. will use this function internally to execute their events.

Let me know if you want to hack on this, otherwise I'm planning on working on it sometime this week or the next.

inversion:
Hi @matthewmueller - I'm currently using it to implement drag and drop as well. (I've made an action which can drag and drop from/to page pixel positions, and you can also pass DOM selectors or functions to evaluate for it to determine those positions).

Internally it uses mouseDown/Move/Up and could also use touchStart/Move/End for mobile simulation.

I'd like to contribute that as I think it could be useful generally. Any thoughts? It would be easy to port without a public sendInputEvent action anyway.

Also just on the upgrade side you'll notice some deprecation warnings with the new Electron which I haven't looked into.

Cheers,
Andrew

matthewmueller:
Wow that's really cool. I'm wondering if that would make more sense as a plugin once we land actions/use.  What do you think?

Thanks for the heads up on the electron warnings!

inversion:
Agreed on the plugin, let's move discussion to #336

maxfriedmann:
Hi @inversion, I stumbled over this issue since I also need some drag and drop functionality. Have you released anything in that direction? Maybe a drag and drop action/plugin?

inversion:
Hi @maxfriedmann - I haven't got anything released for it yet, but the approach I took is basically sending the input events `mouseDown mouseMove (several) mouseUp`. The issues to consider are finding the start and end co-ordinates, choosing where to pick up the element and where to drop it (e.g. if you are trying to drop something that needs to be entirely in a target area, or if there are other target areas adjacent which can be mistakenly dropped into.

PR #363, opened 2015-12-03
scrollTo can be used to scroll to specified selector

Kreozot:
scrollToSelector function + test + readme

### DIFF
--- a/lib/actions.js
+++ b/lib/actions.js
@@ -395,13 +395,13 @@ exports.useragent = function(useragent, done) {
 
 exports.scrollTo = function (y, x, done) {
   debug('.scrollTo()');
-  var selector = undefined;
+  var selector;
   if (typeof x === 'function') {
     done = x;
     selector = y;
   }
 
-  if (selector !== undefined) {
+  if (selector) {
     this._evaluate(function (selector) {
       var element = document.querySelector(selector);
       if (element) {
--- a/lib/actions.js
+++ b/lib/actions.js
@@ -365,6 +365,31 @@ exports.scrollTo = function (y, x, done) {
 };
 
 /**
+ * Set the scroll position to left and top of element by selector.
+ *
+ * @param {String} selector
+ * @param {Function} done
+ */
+
+exports.scrollToSelector = function (selector, done) {
+  debug('.scrollToSelector()');
+  var self = this;
+
+  this._evaluate(function (selector) {
+    var element = document.querySelector(selector);
+    if (element) {
+      var rect = element.getBoundingClientRect();
+      window.scrollTo(rect.left, rect.top);
+    }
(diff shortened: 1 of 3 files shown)


matthewmueller:
Thanks for the PR! Mentioned this in another PR, but I think the API should be:

`nightmare.scrollTo(top, left)`
`nightmare.scrollTo(selector)`

Where scrollTo supports both. What do you think?

Kreozot:
Good point, thank you! Maybe I also should do it in #358, how do you think?

matthewmueller:
@Kreozot haha, yah i was thinking about that... not sure how it'd look unless we break the API. I guess we could maybe do:

```
.screenshot(path, [selector], [, clip])
```

Almost wondering if it just makes more sense to just use an object:

```js
.screenshot({
  path: ...,
  selector: ...,
  clip: ...
})
```

Also, how does this handle multiple elements on different parts of the page by the way?

Kreozot:
It won't breaks the API if we will check argument type.

Kreozot:
@matthewmueller Please take a look.

lib/actions.js
```diff
@@ -386,13 +386,41 @@ exports.useragent = function(useragent, done) {
...
+      var element = document.querySelector(selector);
+      if (element) {
+        var rect = element.getBoundingClientRect();
+        window.scrollTo(Math.round(rect.left), Math.round(rect.top));
+      } else {
+        return 'invalid selector "' + selector + '"';
+      }
+    }, function (a, error) {
+      if (error) {
+        throw new Error(error);
```
matthewmueller: this should probably be: `done(error)`

lib/actions.js
```diff
@@ -386,13 +386,41 @@ exports.useragent = function(useragent, done) {
...
+  }
+
+  if (selector) {
+    this._evaluate(function (selector) {
+      var element = document.querySelector(selector);
+      if (element) {
+        var rect = element.getBoundingClientRect();
+        window.scrollTo(Math.round(rect.left), Math.round(rect.top));
+      } else {
+        return 'invalid selector "' + selector + '"';
```
matthewmueller: i believe the throwing should be done here

lib/actions.js
```diff
@@ -386,13 +386,41 @@ exports.useragent = function(useragent, done) {
...
+      }
+    }, function (a, error) {
```
matthewmueller: what's "a"?

highvoltag3:
@rosshinkley just wondering is there any reason for this PR's to still be awaiting merge?

rosshinkley:
@highvoltag3  I'm still open to this PR, but the source is well out of date and would need to be updated.  Additionally, I don't think @matthewmueller's comments were ever addressed.

PR #365, opened 2015-12-04
alert events

### LINKED ISSUE #359: preload script causes nightmare to hang
If I specify the 'preload' option in nightmare's web-preferences (that get passed through to electron), nightmare hangs. This happens even when the preload script does nothing.

Example script: https://gist.github.com/yoz/b7420bf6fa38d351b8cc
It uses this preload script: https://gist.github.com/yoz/829b05a344d37f3cff26

Expected behavior (which is what we get if the preload line is commented out):
1. open example.com
2. console.log "Example Domain"
3. click a link to http://www.iana.org/domains/reserved
4. console.log "IANA — IANA-managed Reserved Domains"

Actual:
1. open example.com
nothing else happens

rosshinkley:
Sets up the `page-alert` event for alerts, preventing alerts from blocking the process.  Should fix #359.  Thoughts? /cc @yoz

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -186,6 +186,9 @@ This event is triggered if any javscript exception is thrown on the page. But th
 ##### .on('page-log', errorMessage, errorStack)
 This event is triggered if `console.log` is used on the page. But this event is not triggered if the injected javascript code (e.g. via `.evaluate()`) is using `console.log`.
 
+##### .on('page-alert', message)
+This event is triggered if `alert` is used on the page.
+
 #### .screenshot([path][, clip])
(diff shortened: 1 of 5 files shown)


matthewmueller:
sweet! looks great :-D

yoz:
Works great, thanks for the quick update!

yoz:
P.S. window.confirm and window.prompt may also benefit from something like this, although those are more complex because they have to return a value. See for example
http://stackoverflow.com/questions/33520361/nightmarejs-to-click-the-window-confirms-button

rosshinkley:
@yoz I thought about this when I was going through this last night, and decided that I'd address your immediate problem first.  I'm planning to look at confirm and prompt as time permits.

PR #367, opened 2015-12-04
Custom events with bind and unbind

### LINKED ISSUE #354: [question] implement callback function callable within evaluate
Hey there,
is there a way to implement a callback function callable within evaluate?
Like this
```js
nightmare.on('nightmare-message', function(mixed) {
 //do stuff
 console.log(mixed)
})
nightmare
 .goto('example.com')
 .evaluate(function() {
   //do browser stuff
   var myobject = {...}
   //[...]
   window.nightmare(myobject)
   //[...]
   window.nightmare("hello")
 })
```

rosshinkley:
Adds ability to add custom events that can be emitted with `ipc.send` from `evaluate` or `inject`.  Fixes #354.

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -178,13 +178,16 @@ Returns whether the selector exists or not on the page.
 Returns whether the selector is visible or not
 
 #### .on(event, callback)
-Capture page events with the callback. You have to call `.on()` before calling `.goto()`. Supported events are [documented here](http://electron.atom.io/docs/v0.30.0/api/browser-window/#class-webcontents). Additional to the electron-events we provide nightmare-events `'page-error'` and `'page-log'`.
+Capture page events with the callback. You have to call `.on()` before calling `.goto()`. Supported events are [documented here](http://electron.atom.io/docs/v0.30.0/api/browser-window/#class-webcontents). Additional to the electron-events we provide nightmare-events `'page-error'` and `'page-log'`.  Custom events can be added with `.customevent()`.
 
 ##### .on('page-error', errorMessage, errorStack)
 This event is triggered if any javscript exception is thrown on the page. But this event is not triggered if the injected javascript code (e.g. via `.evaluate()`) is throwing an exception.
 
 ##### .on('page-log', errorMessage, errorStack)
 This event is triggered if `console.log` is used on the page. But this event is not triggered if the injected javascript code (e.g. via `.evaluate()`) is using `console.log`.
+
+##### .customevent(name)
+Adds an event that can be captured with `.on()` triggerable in `.evaluate()` or `.inject()` with `ipc.send([name], ...)`.
 
 #### .screenshot([path][, clip])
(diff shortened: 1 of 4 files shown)


matthewmueller:
Sweet! I think the API should be `.bind(...)` though. It's more familiar to folks: 

- http://api.jquery.com/bind/
- https://github.com/component/event

rosshinkley:
I follow and agree.  I will make the update as time permits. :)

rosshinkley:
@matthewmueller I have a couple of questions for you.  One, is it acceptable that `.bind()` accepts _only_ a name, and not data (like say, jQuery's bind)?  Two, `.unbind()` removes all of the named event listeners, introducing the (probably wrong) assumption that unbinding a custom event should remove all of the child process listeners for that name.

Is this closer to what you had in mind?  Thoughts?

matthewmueller:
glad you asked, yep i don't care about the data attribute at all. hmm, what about keeping these separate from `nightmare.on(...)`, where it's like:

```js
nightmare.bind('some-custom-event', function () {

})
```

that would make unbind more clear-cut:

```
nightmare.unbind('some-custom-event', fn) // remove that function
nightmare.unbind('some-custom-event') // remove all functions
```

what do you think?

rosshinkley:
@matthewmueller Sure, making the event handler an option in the function call, or having `unbind` remove the parent listener makes sense.  It's also probably worth pointing out that I don't think you can fully wall this off from `.on()` - the ad hoc events will still be listenable. In other words,

```javascript
nightmare
    .on('some-custom-event', handler)
    .bind('some-custom-event')
```
... should still be valid, if a bit misleading.  Alternatively, and for completeness,

```javascript
nightmare
    .on('some-custom-event', handler)
    .bind('some-custom-event', function(){
        //custom event handler
    })
```
... would also still call the handler from `.on()` when `some-custom-event` is emitted.

It's also probably worth noting that I don't think you can remove an `.on()` handler, either - is there a helper to remove a listener from `child`?

Are you suggesting making the handler function for `.bind()` mandatory?  I'm not so certain about that, but I can't think of a compelling counterargument.

---

What I was a _bit_ more concerned about - and I'm sorry I wasn't terrifically clear - was what happens when you call `.unbind()` down to the child process (in `runner.js`).  I wrote it such that the Electron `.unbind()` listener would scrub all listeners with that name from the renderer, effectively removing that event from being emittable.  This is owing to the handler in the renderer's `.on('some-custom-event', fn)` not being unique and also not being scoped outside of `bind`.

This also raises another question: can more than one event emitter per name be added and bound?  Is there a use case for having multiple custom event emitters of the same event name?  Would it be worth enforcing a rule that one and only one custom emitter could be added per name?

I'm still not sure I'm being clear.  Let me know if you have questions.

rosshinkley:
@matthewmueller I've gone ahead and put in the optional handlers for custom events.  

I've added enforcement of only one child emitter per name.  This prevents situations where custom events can have more than one emitter.  (This also clears up my question from above.)

I've also added some extra logic around how `unbind` works: `.unbind('sample-event')` will remove the event listeners on the parent and the child emitter, where `.unbind('sample-event', handler)` will remove the handler on the parent and the emitter on the child if it is the last handler.  I'm not _real_ confident in the last parent listener removal logic also removing the child emitter.  It's nice the event cleanup is there, but I worry it might be unexpected behavior.  What do you think?

rosshinkley:
Closing this PR.  I will break this out into a plugin pending the outcome of #425 and #391.

PR #368, opened 2015-12-04
Add headers() method, and optional headers param to goto()

### LINKED ISSUE #325: Bump electron and expose extraHeaders to goto?
As of electron 0.34, you can override headers as part of loadUrl: https://github.com/atom/electron/pull/3227

``` javascript
webContents.loadUrl('https://www.example.com', {
  extraHeaders: 'Accept-Encoding: gzip;q=0,deflate;q=0'
});
```

It'd be great if goto could accept a second argument that included these headers. Happy to submit a PR if interested. :)

danielstjules:
Adds a headers() method to override headers for all requests from that instance of nightmare.

``` javascript
yield nightmare
  .headers({ 'X-Nightmare-Header': 'hello world' })
  .goto('localhost:3000/');
```

And adds an optional param to goto(), specifying headers to use for that specific request:

``` javascript
yield nightmare
  .goto('localhost:3000/', { 'X-Nightmare-Header': 'hello world' });
```

Fixes #325

### DIFF
--- a/lib/nightmare.js
+++ b/lib/nightmare.js
@@ -71,6 +71,7 @@ function Nightmare(options) {
   this.ending = false;
   this.ended = false;
   this._queue = [];
+  this._headers = {};
 
   this.child = child(this.proc);
   this.child.once('ready', function() {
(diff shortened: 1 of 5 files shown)


matthewmueller:
thanks @danielstjules! i think the API should be `nightmare.header()` and then support the following:

`header(header, value)`
`header(obj)`

what do you think?

danielstjules:
Sure, I can do that. :) Are you ok with leaving goto's optional headers param as-is?

And any reason for preferring a singular `nightmare.header` as opposed to the original `nightmare.headers` available in Nightmare 1.x? Just trying to keep it more consistent with node's https://nodejs.org/docs/v0.4.0/api/http.html#response.setHeader ?

matthewmueller:
> And any reason for preferring a singular nightmare.header as opposed to the original nightmare.headers

yep and most the APIs that I've seen that have this kind of signature are singular. ex. http://api.jquery.com/attr/

danielstjules:
@matthewmueller Sounds good, updated!

matthewmueller:
thanks @danielstjules !

PR #369, opened 2015-12-05
Viewport size changed in the 'should set viewport' test

Kreozot:
For test passing on small screen resolution.
Because on some platforms electron can not set initial viewport size more than screen resolution.


matthewmueller:
thanks!

Issue #373, opened 2015-12-06
Prevent loading Google Analytics


SerjoPepper:
How can I prevent loading the Google Analytics when I scrape page with nightmare.js?

matthewmueller:
if you're not able to catch the event as the window via inject, it  can be fixed with a custom preload script outlined here: https://github.com/segmentio/nightmare#custom-preload-script

SerjoPepper:
@matthewmueller I expected a more elegant solution, eg blocking scripts loading by domain or hostname, this way works in PhantomJS.

matthewmueller:
let's see if more people need this. there's a million things nightmare *could* do, it's up to us to make sure the API remains simple, while not being limiting.

nquoctuan:
Hi @matthewmueller ,

I am not really sure the issue that I create a custom preload script to capture alert message. Actually I follow the site that you provided, but no chance.

```
VM106:3 Uncaught TypeError: Cannot read property 'nightmare' of undefined
javascript @ VM106:3
(anonymous function) @ VM106:38
electron.ipcRenderer.on @ init.js:51
emitMany @ events.js:127
emit @ events.js:201
 ```

**Comment by nquoctuan** [comment]
source code is quite simple:
```
var Nightmare = require('nightmare'),
  nightmare = Nightmare({
    openDevTools: {
      mode:'detach'
    },
  show: true,
  webPreferences: {
        devTools: true,
        preload: 'alertMessage.js',
        nodeIntegration: false,
        webSecurity: false,
        allowRunningInsecureContent: true
      }
});
  
nightmare.useragent('Mozilla/5.0 (Macintosh; Intel Mac OS X 10_11_6) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/58.0.3015.0 Safari/537.36');

nightmare.goto('http://www.google.com')
  .then(function() {
	return nightmare.url();  
  })
  .then(u => {
	  console.log('The url is: ' + u);
  });
```

rosshinkley:
@nquoctuan I'm _almost_ positive that the path for `preload` has to be fully qualified.  What happens if you change that to `preload: require('path').resolve('alertMessage.js')`?

Issue #366, opened 2015-12-04
preload template


rosshinkley:
@yoz asked a good question about overriding preload functionality in #359, and as an extension, I have a question of my own: Is there a use case for adding custom parts inside of Nightmare's preload?

Preload could be a template, but I think this would require writing the compiled preload script to disk (Electron expects a file path for preload, at least in my read of the documentation).  Electron exposes the default paths for user data, and changes made in #338 allows for custom paths.  Both could be used to determine a place to save the compiled preload.

Thoughts?

_Edit:_ Bah, a second look at the mentioned PR and at Electron's documentation, I saw `.getPath(name)` almost right away.  Edited commentary.

rosshinkley:
Moved conversation to #374.  Closing.

Issue #364, opened 2015-12-03
use standard


matthewmueller:
use husky's precommit hook + standard to help resolve formatting issues

PR #374, opened 2015-12-07
blocking prompts, preload additions

rosshinkley:
This takes care of @yoz's question from PR #365 regarding `confirm` and `prompt`: by default, they will both emit their respective events and throw an exception.  This prevents unwanted behavior on the page (eg, by prompt/confirm returning default values), as well as prevents Nightmare from hanging on modal dialogs.

This also is a partial answer to my question #366.  Rather than overriding the Nightmare preload, the internal preload script can now be added to with the same `options['web-preferences'].preload` option.  This allows for overriding the above exception behavior if it is not desirable, as well as opens up user-defined preload actions should they be needed.

Thoughts?


test/index.js
```diff
@@ -517,6 +518,102 @@ describe('Nightmare', function () {
       alert.should.equal('my alert');
     });
 
+    it('should fire an event and except on javascript window.prompt', function*() {
```
yoz: s/except/accept/

test/index.js
```diff
@@ -517,6 +518,102 @@ describe('Nightmare', function () {
...
+      } catch(e) {
+        didFail=true;
+      }
+
+      didFail.should.be.true;
+      prompt.should.equal('my prompt');
+      defaultResponse.should.equal('default value');
+    });
+
+    it('should fire an event and except on javascript window.confirm', function*() {
```
yoz: same here

Readme.md
```diff
@@ -115,6 +115,9 @@ var nightmare = Nightmare({
 });
 ```
 
+##### ['web-preferences'].preload
+Specifies a file to add to preload.  The generated preload file is written to Electron's `userData` path.
```
yoz: This isn't particularly clear, plus I think a user of this doesn't need to know where the generated file goes. How about:
Specifies a file to add to Electron's `preload`. (The generated preload file is written to Electron's `userData` path.)

Readme.md
```diff
@@ -198,6 +201,12 @@ This event is triggered if `console.log` is used on the page. But this event is
 ##### .on('page-alert', message)
 This event is triggered if `alert` is used on the page.
 
+##### .on('page-prompt', message [, response])
+This event is triggered if `prompt` is used on the page.  Note that unless the `preload` script is overridden, the prompt will cause an exception to be thrown.
```
yoz: Can this say:
Note that unless `window.prompt` is overridden in a preload script, ...

(and similarly below for page-confirm)

Readme.md
rosshinkley: I agree this line isn't particularly clear, and I struggled with making something concise.

I think "...add to Electron's `preload`" is a bit imprecise as the user's preload is being spliced into the Nightmare preload.  I wanted to give enough information to make it clear that the file provided was not the file used, and where it was saved to on disk without getting into too much detail.  

How about something like:

>Specifies a file to add to Nightmare's preload.  The generated preload file is written to Electron's `userData` path.

Readme.md
rosshinkley: Updated. :)

Readme.md
yoz: Sure, I think I still like the parentheses around the second sentence though to indicate that it's a separate note of lesser importance.

matthewmueller:
Hey thanks for the PR! I think we should actually group these events: https://github.com/segmentio/nightmare/pull/337

I should have considered this before accepting the alert one, but I'm thinking something along the times of: 

`nightmare.on('page', function (type, message[, response])`

---

As far as the preloading stuff, I don't think it's a great idea to completely allow you to overwrite it because as we make changes to the API, those people that overwrite it will be completely out of sync. Are there specific use-cases that we should allow custom hooks into the preload script?

rosshinkley:
@matthewmueller Ah, yeah, grouping them makes a lot of sense.  I'll pull that together as time permits.

As for overwriting the preload: the template-generated preload would be overwritten every time.  The template is internal to Nightmare.  As such, the internal changes to preload should be picked up if any are made.  Ultimately, what I am reaching for is the default exception behavior for prompt and confirm may be (probably is?) undesirable, and there should be a way to override it.  Maybe I'm not following what you're asking?

As for specific use cases: the only one I know of offhand is what @yoz ran into with `alert` - if you have a prompt or confirm an event that fires _before_ an `inject`/`evaluate` can run.  This gets back to overriding the default exception behavior - I'd imagine you'd want to overwrite it to handle dialogs with valid responses.  Does this make sense?

Readme.md
rosshinkley: Added.  Thanks, @yoz.

rosshinkley:
@matthewmueller I may have gotten a _bit_ overzealous (and these changes almost certainly step on #337's logging changes), but I've grouped all of the page-* events into one unified event.  I also tried to update the documentation to be clear-ish on what to expect with `on('page', type ...` using the jsdoc-like format.  (Offtopic, is there a documentation standard to follow?  Is this tangentially related to #364?)

Thoughts?

matthewmueller:
Beautiful. I like this a lot. I like how you were able to consolidate all these into `page`. I think what you did with the custom preload script is good too. Two things I'd like to explore:

1. Is it possible to maintain the syntax highlighting on the preload script? I've noticed it's pretty error-prone working with functions as strings.
2. Is it possible to have the custom preload script to be a function that passes in the IPC

```js
function (ipc) {
  window.confirm = function(message) {
      var response = false;
      if(message == 'foo'){
          response = true;
      }
      ipc.send('page', 'confirm', message, response);
      return response;
  };
}
```

That might look be a little nicer way to extend the preload script and also authors don't need to know anything about the internal __nightmare.ipc

rosshinkley:
@matthewmueller thank you. :)  Responding to your questions:

1. Yeah, the other reason I went the templated route was because of precedent, but regardless, this is a good question.  Spitballing an idea off the top of my head: could the internal preload script be in its own file, read off disk, have the `{{!src}}` token spliced in, and have that be used in the place of `exports.preload`?  That would let both the user-defined preload and internal preload be real first-class javascript files, complete with syntax highlighting.  (There are probably less ugly, less fragile variants of this, like having the token in a comment or something.  The chosen solution might be able to be adapted for the other templates.  Would that worth pursuing in a separate PR?)
1. Thinking about this with a little distance, I could have probably simply used `require` to pull in the user-defined preload, and pass the IPC to it that way.  Should the token then be changed to something like `require('{{!userPreloadFilepath}}')(__nightmare.ipc);`?  I'm almost positive the path would _have_ to be absolute, and the documentation should reflect that.  This also may help solve the token issues from your first question if the require had a dummy file to require by default.  Thoughts?

rosshinkley:
@matthewmueller This was an itch I had to scratch: I went ahead and pulled together the `require` approach described above.  Is this close to what you had in mind?

rosshinkley:
By the by, back to the earlier topic of use cases: Providing custom preload scripts should let you prevent elements from rendering on a page, and as such, I think this should provide a way to fix #373.

lib/runner.js
```diff
@@ -49,11 +51,17 @@ app.on('ready', function() {
    */
 
   parent.on('browser-initialize', function(options) {
+    var userPreload = resolve('userPreloadTemplate.js');
```
matthewmueller: mind if we rename this to `user-preload-template.js`? not a huge deal, but i try to keep everything lowercase to avoid any issues case-sensitive filesystems.

test/files/preload.js
```diff
@@ -0,0 +1,21 @@
...
+        __nightmare.ipc.send('page', 'prompt', message, response);
+        return response;
+    };
+
+    window.confirm = function(message) {
+        var response = false;
+        if(message == 'foo'){
+            response = true;
+        }
+        __nightmare.ipc.send('page', 'confirm', message, response);
```
matthewmueller: we should update this test to reflect the new `require('{{!userPreloadFilepath}}')(__nightmare.ipc);`

lib/runner.js
matthewmueller: wait, actually what does this do? is the `userPreloadTemplate.js` where you'd put your custom implementation?

lib/preloadTemplate.js
```diff
@@ -0,0 +1,39 @@
+/**
```
matthewmueller: same deal as `userPreloadTemplate`, please rename to `preload-template.js`

lib/runner.js
rosshinkley: The intent here was to provide a file that was guaranteed to exist so the `require` call in preload wouldn't cause an exception.  Wrapping it in a try/catch and eating the exception felt incorrect, but I have no real facts to back that up.  Do you have a preference?  

The custom implementation for preload can be at any filepath defined in `options.preloadPath`.

That being said, this really ought to use `join` instead of `resolve`.  Fix coming.

test/files/preload.js
rosshinkley: Absolutely, this got overlooked.  Fix coming.

lib/runner.js
rosshinkley: Adding a fix to do something akin to...

```javascript
if('{{!userPreloadFilepath}}' != ''){
    require('{{!userPreloadFilepath}}')(__nightmare.ipc);
}
```

... removes the need for the placeholder or try/catch.  I didn't like this at first because it makes the generated file logic look... awkward?  but the more I think about it, the better it seems.

matthewmueller:
Thanks for this PR! I've incorporated these changes into master!

matthewmueller:
One thing I left out was the preload templating stuff (probably due to my suggestion :-/), but I think it's less complicated to just include your own script

rosshinkley:
Yeah, I understand what you did in your PR.  I would have preferred skipping the user having to include the IPC themselves, but what you did works just fine too. :)  Thanks!

matthewmueller:
yah me too :/, but the added complexity wasn't worth it imo

Issue #371, opened 2015-12-05
Enable adding event listeners to webContents.session


pkawiak:
When I ```goto``` a web page containing ```iframe``` with a pdf, save file dialog pops up blocking the process. I've found out that it is possible to resolve that by adding ```will-download``` listener to ```webContents.session``` of ```BrowserWindow``` as described [here](https://github.com/atom/electron/issues/3142). Would it be desirable to extend Nightmare's API to support this usecase? If yes then would sth like that be fine?
```
nightmare.session.on('will-download', handler);
```
If that's acceptable then I could submit PR

rosshinkley:
@fmaruki, @darthcloud and myself touched on this in #332.  I didn't pull together a PR because I wasn't too keen on polluting options with download rules and I was/I am _pretty_ certain my implementation and approach were not good.  Since then, I haven't set aside time to take a second (hopefully longer) look.

Adding an event handler in this way might work.  It side-steps the immediate action problem, and offloads responsibility for cleanup and multiple downloads to the user, which simplifies things.

Thoughts?

pkawiak:
Is there any other way we can automate pages containing iframes with pdf content? I don't even want to download it, I just don't want it to get in my way by displaying blocking system window.

Would it be acceptable to just hardcode ```will-download``` event prevention in Nightmare?

Is it of any use to display blocking system dialog in context of browser automation? If yes then is there any way to interact with it in automated way?

rosshinkley:
While I understand the temptation to discard/ignore downloads, I don't think that is a particularly good approach.  If Nightmare is being used for scripting, it is possible that the downloaded file will help determine what steps to take after the download (eg, with integration testing).  I suppose Nightmare could simply accept every download, but that also seems like an unpleasant option.

Answering your second question: From within browser automation, no, modal system dialogs are not useful and cause problems.  For example, `alert`, `confirm`, and `prompt` cause Nightmare to hang just like with download prompts.  (For what it's worth, some work has been done to address those three, and for more information, see #359, #365, #366 and #374.)

Answering your third question: from within Electron, you can either cancel the download or set the download path within the `will-download` handler.  This is functionally the same as interacting with the download dialog and will prevent one from showing.

pkawiak:
[...]

PR #382, opened 2015-12-11
handle will-download

### LINKED ISSUE #151: Cannot download a file
Has anyone been able to download a file?

I have tried a few approaches. The first was:

```js
.on('filePicker', function(oldFile) {
            console.log('onFilePicker(' + oldFile + ') called');
            return 'a-filename.ext';
          })
```

That code was never triggered.

Then [someone mentioned](https://groups.google.com/forum/#!topic/phantomjs/JChUakj--24) onFileDownload, but that also didn't work.

I have a listener to output resource requests (and it is correctly logging the attempt to retrieve the file):

```js
.on('resourceRequested', function(requestData, networkRequest) {
            console.log('Request (#' + requestData.id + '): ' + JSON.stringify(requestData));
          })
```

I can't just stream the file to disk because it's behind an authenticated session, hence the need for PhantomJS.

Anyone able to get file downloading to work? I think [Casper](http://docs.casperjs.org/en/latest/modules/casper.html#casper-download) has it going, so I will try that next.

### LINKED ISSUE #332: Prevent automatic downloads
When visiting a page that prompts a download window after the onload, is there a way to prevent it to pop up, so that I can use nightmarejs without needing to interact and close those windows?

Something like a preventDefault on a will-download event?

rosshinkley:
Allow users to automatically accept, automatically ignore, or handle downloads on a per-case basis. Fixes #332 and #371 with the addition of the `ignoreDownloads` option, should fix #151 with the `download` event and the various ways that can be used.

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -116,6 +116,12 @@ var nightmare = Nightmare({
 });
 ```
 
+##### ignoreDownloads
+Defines whether or not downloads should be ignored.
+
+##### downloadTimeout
+This will throw an eception if the `.wait('downloads-complete')` didn't return `true` within the set timeframe.
+
 #### .useragent(useragent)
 Set the `useragent` used by electron.
 
@@ -178,6 +184,11 @@ Wait until the element `selector` is present e.g. `.wait('#pay-button')`
 #### .wait(fn[, arg1, arg2,...])
 Wait until the `fn` evaluated on the page with `arg1, arg2,...` returns `true`. All the `args` are optional. See `.evaluate()` for usage.
 
+#### .wait('downloads-complete')
+Wait until all downloads are in a state of `'completed'`, `'interrupted'`, or `'cancelled'`
+
+#### .emit(eventType[, arg1, arg2,...])
+Sends an event of `eventType` to the Electron process.
 
 ### Extract from the Page
 
@@ -198,6 +209,30 @@ This event is triggered if `console.log` is used on the page. But this event is 
 
 ##### .on('page-alert', message)
 This event is triggered if `alert` is used on the page.
+
+##### .on('download', event, downloadItem)
+This event is triggered when Electron emits `'will-download'`.  This event is also emitted after downloads are started when [`DownloadItem`](https://github.com/atom/electron/blob/master/docs/api/download-item.md) emits `'updated'` or `'done'`.   The possible values for `event` are `'started'`, `'cancelled'`, `'interrupted'`, or `'completed'`.  Note that by listening to `'download'`, Nightmare expects the default download behavior to be overridden. 
+
+Downloads should be completed using `Nightmare.emit('download', action, downloadItem)`.  The possible values for `action` are `'cancel'`, `'continue'` for default behavior, or a file path (file name and extension inclusive) to save the download to an alternative location. The `downloadItem` parameter should use the item passed by `'download'`.
+
+For example:
+
+```javascript
+var nightmare = Nightmare();
+nightmare.on('download', function(event, downloadItem){
+  if(state == 'start'){
+    nightmare.emit('download', '/some/path/file.zip', downloadItem);
+  }
+});
+
+yield nightmare
+  .goto('https://github.com/segmentio/nightmare')
+  .click('a[href="/segmentio/nightmare/archive/master.zip"]')
+  .wait('downloads-complete');
+```
+
+By default, downloads are saved to the path defined in `paths.downloads`.  Also by default, downloads are automatically accepted (`nightmare.emit('download', 'continue', downloadItem)` is performed automatically internally in Nightmare) unless `ignoreDownloads` is specified.
+
 
 #### .screenshot([path][, clip])
 Takes a screenshot of the current page. Useful for debugging. The output is always a `png`. Both arguments are optional. If `path` is provided, it saves the image to the disk. Otherwise it returns a `Buffer` of the image data. If `clip` is provided (as [documented here](https://github.com/atom/electron/blob/master/docs/api/browser-window.md#wincapturepagerect-callback)), the image will be clipped to the rectangle.
--- a/Readme.md
+++ b/Readme.md
@@ -120,7 +120,7 @@ var nightmare = Nightmare({
 Defines whether or not downloads should be ignored.
 
 ##### downloadTimeout
-This will throw an eception if the `.wait('downloads-complete')` didn't return `true` within the set timeframe.
+This will throw an exception if the `.wait('downloads-complete')` didn't return `true` within the set timeframe.
 
 #### .useragent(useragent)
 Set the `useragent` used by electron.
@@ -219,8 +219,8 @@ For example:
 
 ```javascript
 var nightmare = Nightmare();
-nightmare.on('download', function(event, downloadItem){
-  if(state == 'start'){
+nightmare.on('download', function(state, downloadItem){
+  if(state == 'started'){
     nightmare.emit('download', '/some/path/file.zip', downloadItem);
   }
 });
--- a/lib/actions.js
+++ b/lib/actions.js
@@ -233,8 +233,13 @@ exports.wait = function () {
     waitms(arg, done);
   }
   else if (typeof arg === 'string') {
-    debug('.wait() for ' + arg + ' element');
-    waitelem(this, arg, done);
+    if(arg == 'downloads-complete'){
+      debug('.wait() for downloads');
+      waitDownloads(this, done);
+    } else {
+      debug('.wait() for ' + arg + ' element');
+      waitelem(this, arg, done);
+    }
   }
   else if (typeof arg === 'function') {
     debug('.wait() for fn');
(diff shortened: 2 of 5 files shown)


Readme.md
```diff
@@ -199,6 +210,30 @@ This event is triggered if `console.log` is used on the page. But this event is
...
+This event is triggered when Electron emits `'will-download'`.  This event is also emitted after downloads are started when [`DownloadItem`](https://github.com/atom/electron/blob/master/docs/api/download-item.md) emits `'updated'` or `'done'`.   The possible values for `event` are `'started'`, `'cancelled'`, `'interrupted'`, or `'completed'`.  Note that by listening to `'download'`, Nightmare expects the default download behavior to be overridden. 
+
+Downloads should be completed using `Nightmare.emit('download', action, downloadItem)`.  The possible values for `action` are `'cancel'`, `'continue'` for default behavior, or a file path (file name and extension inclusive) to save the download to an alternative location. The `downloadItem` parameter should use the item passed by `'download'`.
+
+For example:
+
+```javascript
+var nightmare = Nightmare();
+nightmare.on('download', function(event, downloadItem){
+  if(state == 'start'){
```
wcandillon: the state variable doesn't seem to be defined in this example.

Readme.md
```diff
@@ -116,6 +116,12 @@ var nightmare = Nightmare({
 });
 ```
 
+##### ignoreDownloads
+Defines whether or not downloads should be ignored.
+
+##### downloadTimeout
+This will throw an eception if the `.wait('downloads-complete')` didn't return `true` within the set timeframe.
```
wcandillon: exception*

wcandillon:
I reviewed it and tested it. It works great. Thank you so much for doing this.

Readme.md
```diff
@@ -178,6 +184,11 @@ Wait until the element `selector` is present e.g. `.wait('#pay-button')`
 #### .wait(fn[, arg1, arg2,...])
 Wait until the `fn` evaluated on the page with `arg1, arg2,...` returns `true`. All the `args` are optional. See `.evaluate()` for usage.
 
+#### .wait('downloads-complete')
+Wait until all downloads are in a state of `'completed'`, `'interrupted'`, or `'cancelled'`
```
matthewmueller: i don't think it's a good idea to overload `nightmare.wait(str)` in this way. It sounds great semantically, but if we have custom tags like `<download-complete></download-complete>`, we'll run into problems.

rosshinkley: Fair enough.  What do you propose as an alternative?  Should there be a standalone method, eg `.waitForDownloads()`, or should the wait string start with a character that is illegal for tags, eg `.wait(':downloads-complete')`?

Readme.md
rosshinkley: Good eye, thank you. :)

Readme.md
rosshinkley: Ah, man, I missed the rename.  It should be `state == 'started'`, with `event` renamed to `state`.  The event name is `download`, and I wanted to be clear about what the first parameter was.

Also, `'start'` should be `'started'`.  Fix coming.  Thanks for the find!

matthewmueller:
thanks for putting this together and i appreciate all the work that was put into this. currently this approach adds too much complexity to the API for solving this issue. 

i've thought about this a little bit and i think a better approach might be:

1. By default ignore all downloads
2. If someone wants to download something, provide `nightmare.download(selector|url)` which returns a buffer. if a buffer is not possible, support `path` in the signature to write to a file.
3. When we are downloading something, wait until that download is complete in order to resume.
4. Potentially support progress events via `nightmare.on('progress', fn)` if that's supported

This increases the API to:

```js
nightmare
  .on('progress', fn)
  .download('a[href="/segmentio/nightmare/archive/master.zip"]')
```

One thing we should also consider is... does nightmare even need to support downloads, or can we get away with:

```js
var request = require('request')
var fs = require('fs')

var url = yield nightmare
  .evaluate(function() {
    return document.querySelector('a[href="/segmentio/nightmare/archive/master.zip"]').getAttribute('href')
  })

request.get('url').pipe(fs.createWriteStream(__dirname + 'somepath.txt'))
```

Are there any advantages to having this in core?

subcontact:
Firstly - awesome library and great pull request.
I wanted to share a scenario where I am logged in with a session and need to click on a button that runs some javascript that POSTS a form with a unique URL and data set and cannot only be requested once (each submission creates a new ID at the end of the URL so cannot be replayed).
The response is a download attachment (Content-Disposition: attachment).

I need to get get the response attachment somehow - an API would allow for that. Currently I haven't figured out how to do this.

Thanks

matthewmueller:
@subcontact hmmm.. that's a tough one, have you tried getting it to work with the current PR?

rosshinkley:
@matthewmueller Thank you for your reply, and I'm sorry I couldn't get back to this sooner.  Also, apologies in advance for the missive.  Responding to your feedback:

> By default ignore all downloads

Seems reasonable to me.  I can do that as time permits.

> If someone wants to download something, provide nightmare.download(selector|url) which returns a buffer. if a buffer is not possible, support path in the signature to write to a file.

I don't understand what you're asking, what should the selector provided do?  Click?  As for providing a URL, would that be a simple get?  And with the buffer, no, I don't think that is possible.  Electron supplies the filepath and a method to override the filepath.

> When we are downloading something, wait until that download is complete in order to resume.

Blocking on each download isn't how browsers work.  From a test automation standpoint, I can certainly understand the need, but from a purely automation standpoint, I'm not sure I agree.

> Potentially support progress events via nightmare.on('progress', fn) if that's supported

Yes, that's supported (and can be supported with the reduced API).

Re: getting away with a simple request: no, I don't think you want to do that.  If you're making an authenticated request for a resource (thinking of something like a request to a private link in say, Dropbox), you want to do that from within the context of Electron.  Copying the resource URI and issuing a plain GET will almost certainly result in failure.  This (I think, and I'm sorry if I'm misreading here) gets to what @subcontact's problem - authenticated downloads won't work with simple requests.

Re: keeping at least some form of downloads in core, that's a _great_ question.  My answer to that would be a qualified no.  How downloads are handled is probably going to be dependent on use case.  To that end, I think the blocking `.download()` and the non-blocking methods in this PR should be broken out into at least one (probably two?) plugins.  

The problem with that is binding to `will-download`.  A new method to hook into `will-download` with a custom handler could be exposed, or (with modification to pass a child handler) the custom event bindings in #367 could be used.  Thoughts?

I'm willing to reduce the API as requested, and if you think it's a good idea, I will close this PR and open a fresh one with those changes.

matthewmueller:
@rosshinkley thanks for the detailed response!

> I don't understand what you're asking, what should the selector provided do? Click? As for providing a URL, would that be a simple get? And with the buffer, no, I don't think that is possible. Electron supplies the filepath and a method to override the filepath.

Yah, you're right, when I proposed that I was thinking about it in terms of what the click/mouseover/etc methods do, but that's not the same as download. 

> Blocking on each download isn't how browsers work. From a test automation standpoint, I can certainly understand the need, but from a purely automation standpoint, I'm not sure I agree.

I was proposing this to handle `wait('downloads-complete')` piece that you added in this PR. Haven't investigated how that would work internally though.

> Re: getting away with a simple request: no, I don't think you want to do that. If you're making an authenticated request for a resource (thinking of something like a request to a private link in say, Dropbox), you want to do that from within the context of Electron. Copying the resource URI and issuing a plain GET will almost certainly result in failure. This (I think, and I'm sorry if I'm misreading here) gets to what @subcontact's problem - authenticated downloads won't work with simple requests.

From my understanding, a POST request triggers the download, so whether this is a plugin or not, we'd need to at least know how to do that with core.

> Re: keeping at least some form of downloads in core, that's a great question. My answer to that would be a qualified no. How downloads are handled is probably going to be dependent on use case. To that end, I think the blocking .download() and the non-blocking methods in this PR should be broken out into at least one (probably two?) plugins.

Yah, I agree.

> The problem with that is binding to will-download. A new method to hook into will-download with a custom handler could be exposed, or (with modification to pass a child handler) the custom event bindings in #367 could be used. Thoughts?

Yah, I think we should add this regardless, we're already hooking into the other events. Seems like it would be as simple as:

`win.webContents.session.on('will-download', forward('will-download'))`

Then you can hook into it with an action:

```js
Nightmare.action('download', function () {
  this.child.on('will-download', handler)
})
```

Haven't thought through exactly how this would work, but that would be the approach I would investigate.

rosshinkley:
@matthewmueller no problem. :)  Thanks for yours as well!

Tying up some loose ends:

> I was proposing this to handle `wait('downloads-complete')` piece that you added in this PR. Haven't investigated how that would work internally though.

Yeah, I understand the need, and also buy `.wait('downloads-complete')` is problematic.  Totally off the cuff: I think that enforcing one download per `.download()` call would probably work.  It would consist of (more or less) how it works now, but with an extra wait for `will-download` to be caught, and the internal method using the wait-download function.

> From my understanding, a POST request triggers the download, so whether this is a plugin or not, we'd need to at least know how to do that with core.

I was more concerned with the "authenticated" part, not with that it was triggered on POST.  I don't _think_ it should matter.  Electron doesn't seem to care what triggered a download, just that one happens.  That being said, it's probably worth propping up a minimal test case to be certain.

> Yah, I agree [to plugins].

The more I think about this, the more sense it makes, and maybe should redirect my attention to #391.  I was thinking about how to extend the Electron process earlier today, because, well...

> Seems like it would be as simple as: `win.webContents.session.on('will-download', forward('will-download'))`

Handling this appropriately seems to require doing some special steps in the Electron process.  In tinkering with `.on('will-download', fn)`, as soon as the `DownloadItem` fell out of the handler scope, Electron would execute whatever behavior had been defined by synchronous completion (almost like the `on` handler is more like middleware, if that makes sense).  The [Electron docs for `setSavePath`](https://github.com/atom/electron/blob/master/docs/api/download-item.md#downloaditemsetsavepathpath) seem to support what I found by experimentation:

>The API is only available in session's will-download callback function. If user doesn't set the save path via the API, Electron will use the original routine to determine the save path(Usually prompts a save dialog).

In other words, trying to do something via event communication would (almost) certainly result in a blocking download dialog.

I think I'm going to close this PR, backburner the download plugins for the moment, and direct my remaining questions to #391 as I think they're more sensible under that topic.

AspireToCodeBetter:
@rosshinkley since the merge of this PR seems to be on hold. would you be able to provide some example to download a file using the internals? Sorry, I am new to both electron and nightmare

rosshinkley:
@AspireToCodeBetter TL;DR: At present, you'd have to fork Nightmare and roll your own.

There is currently work being done to add extensions to the Electron process that Nightmare wraps (#391), one of the benefits of this being the ability to attach a handler to `will-download`.  The reason this PR was ultimately closed is because there are several use-cases for `will-download`, leading me to believe they should be broken out into plugins to be used with (as of this writing, incomplete) extending Electron.

When #391 is complete, I'll see to it that the download plugins are completed and put on NPM, and  certainly post a link on this thread and possibly on a few others (#151 and #371 come to mind).

PR #383, opened 2015-12-13
Added 'electronPath' option to customize Electron path

rhysd:
I added `electronPath` option to options of `Nightmare()`.

I added this because I want to use Nightmare for end-to-end Electron app testing (e.g. Nightmare uses Electron v0.35.2 but my app uses new API introduced on Electron v0.35.4).  Nightmare now fixes Electron version.  I agree that it's reasonable bacause Nightmare can't support all versions of Electron.  So I added notation to document.

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -116,6 +116,15 @@ var nightmare = Nightmare({
 });
 ```
 
+#### electronPath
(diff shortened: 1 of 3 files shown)


matthewmueller:
thanks!

PR #385, opened 2015-12-13
Followed new BrowserWindow option naming

rhysd:
Option naming of `BrowserWindow` was changed at Electron v0.35. (e.g. `max-width` -> `maxWidth`)

Documentation:
https://github.com/atom/electron/blob/master/docs/api/browser-window.md#new-browserwindowoptions

### DIFF
--- a/test/index.js
+++ b/test/index.js
@@ -715,8 +715,8 @@ describe('Nightmare', function () {
       headers['x-nightmare-header'].should.equal('hello world');
     });
 
-    it('should allow web-preferece settings', function*() {
-      nightmare = Nightmare({'web-preferences': {'web-security': false}});
+    it('should allow webPrefereces settings', function*() {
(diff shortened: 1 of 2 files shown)


test/index.js
```diff
@@ -715,8 +715,8 @@ describe('Nightmare', function () {
       headers['x-nightmare-header'].should.equal('hello world');
     });
 
-    it('should allow web-preferece settings', function*() {
-      nightmare = Nightmare({'web-preferences': {'web-security': false}});
+    it('should allow webPrefereces settings', function*() {
```
matthewmueller: looks like we made this error before but do you mind correcting this? `webPrefereces => webPreferences`?

matthewmueller:
thanks! will change that comment on my end :-)

rhysd:
@matthewmueller Thanks!

PR #387, opened 2015-12-15
PDF Options Out of Date

jlukic:
I've noticed that you're bundling `0.35.2` which includes additional options for `printToPDF`, including additional paper sizes like `letter`, `A5` etc. 

The current documentation points to an older version of Electron docs which do not show all possible settings. This PR updates the link to reference the docs specifically for `0.35.2`


matthewmueller:
awesome, thanks!

PR #399, opened 2015-12-21
clear input on falsey/empty values

### LINKED ISSUE #394: .type() behavior changed in 2.1.0
After updating to nightmare@2.1.0 the way the `.type()` method works seems to have changed. In 2.0.9 and before, `.type()` would clear out the element and then enter the specified text, but after updating to 2.1.0, the specified text is typed in, but the existing text is not cleared out. Is this the desired outcome or a bug? If this is the desired behavior, is there a way to delete the text inside an element? With 2.0.9 I could just do `.type(selector, '')` to clear the text from an element, but now that will just add an empty string to the beginning of the value already there.

rosshinkley:
Fixes #394.

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -151,8 +151,8 @@ Clicks the `selector` element once.
 #### .mousedown(selector)
 Mousedown the `selector` element once.
 
-#### .type(selector, text)
-Enters the `text` provided into the `selector` element.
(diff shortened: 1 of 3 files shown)


matthewmueller:
awesome, thanks @rosshinkley!

PR #401, opened 2015-12-22
Blur focussed on click.  Fixes #400

### LINKED ISSUE #400: click() does not unfocus other screen elements
As I have event handlers that update state when a text field is blurred, I must have a way to blur the element, and clicking on a button should do that.

naturalethic:


naturalethic:
I switched it to use activeElement and added a unit test.

matthewmueller:
thanks @naturalethic !

Issue #282, opened 2015-10-12
clean up after event listeners


matthewmueller:
if you're running these things concurrency, it's pretty easy to run into:

```
(node) warning: possible EventEmitter memory leak detected. 11 uncaughtException listeners added. Use emitter.setMaxListeners() to increase limit.
```

souly1:
+1 for the issue

LarryBattle:
@matthewmueller Could you provide a test example?

FredXue:
in my case, after adding a .wait(selector) action into the test, it runs into this situation. Any idea? or solution?

LarryBattle:
The fix/hack is to add `process.setMaxListeners(0);` at the top of your script.
Source: http://stackoverflow.com/questions/9768444/possible-eventemitter-memory-leak-detected

To reproduce the warning message, remove `process.setMaxListeners(0);` from `test/index.js` then run the tests.

I wonder if there is actually a memory leak going on.

philhaz:
process.setMaxListeners(0); is making no difference here - definitely seems related to the use of wait().

mocha 2.3.3, node 4.1.2, nightmare 2.0.7

pcalin:
Same problem here. process.setMaxListeners(0) did not make a difference.

nightink:
+1 for the issue

victorwpbastos:
+1 for the issue

fr-:
It seems to me that using nightmare in a long running/concurrent setup is prone to many errors and can cause all kinds of hiccups.

Firstly, there are _some_ things to consider when registering a global (process) event handler in the ctor of any class, but in this specific case it's pretty important to clean them up in _all_ cases that can be considered a shutdown/dtor scenario for nightmare: 

1. Shutdown via end()
2. Exit of child process
3. Inability to start child process
4. Exceptional exit

This also applies for some of the control flow, which can cause nightmare to hang if the electron process abnormally quits, as it does not reject in those cases (mainly never returning from ``continue``).

I have created a preliminary feature branch on my fork of nightmare that tries to tackle these, but it's not cleaned up. Basically it tries to achieve multiple things:

1. Clean up after event listeners
2. Dynamically increase the maximum event listeners when more instances of nightmare are created on a single process
3. Handle abnormal exists (such as electron SIGSEGV/SIGBUS, exit code != 0, exceptions)
4. Set a ``failed`` state on the nightmare instance along the lines of ``initial`` and ``ready`` in case electron crashes/exits abnormally

[...]

PR #407, opened 2015-12-23
fix eventemitter leak message

contra:
if you spin up 10 nightmare processes, you get nasty console messages about an event emitter leak because 10+ listeners will be attached to process. this tells node everything is okay and we meant to do this.


matthewmueller:
awesome, thanks. we still need to cleanup after our emitters better #282, but at least we won't get that annoying warning all the time.

jimthedev:
Any idea when this will make it in?

Issue #414, opened 2015-12-26
Add more reliable way to populate a text into an element


nodaguti:
I found that `nightmare.type()` does not wait for actually finishing the typing since v2.1.0 (i.e. starting to use `sendInputEvent`), which causing populating a long text into an input element to fail in an certain situation.
I created a sample application to describe this issue:

https://github.com/nodaguti/nightmare-type-issue

In this app, `click('[type=submit]')` is executed before finishing to type a long message, resulting that only a partial message is submitted.

This issue is annoying especially when you test a login page or something like that because you have to put `wait(100)` before every `type()` actions to ensure the entire user ID and password are populated.

`type()` is still useful to trigger keyboard-related events like `keypress` properly, so I suggest adding another action `populate(selector, text)` to enter a text into an element synchronously by executing `element.value = text`.

I also created a pull request to implement `populate` action.

matthewmueller:
Let's track this issue: https://github.com/atom/electron/issues/3930

nodaguti:
Thanks for creating an issue for Electron about this problem!
Indeed the root of this issue lies `sendInputEvent` 's asynchoronous, so it's perfect if it is resolved on Electron side.

matthewmueller:
Okay this is turning into a pretty tricky issue. According to #3930, there's not a good way to track asynchronous events. Right now I'm considering two options but could use some feedback.

1) Send less events. Programmatically set the value, then trigger the last character's events (like you're just finishing typing something)

2) Keep current functionality. I've tested it manually, and I haven't been able to reproduce this issue on smaller bodies of text (yet), so it at least it seems somewhat rare. Add an point in an FAQ on how to either override or add a custom action like populate to nightmare (via `.action()`)

I'm leaning towards 1), but I could use some feedback :-)

stephenmathieson:
>  I haven't been able to reproduce this issue on smaller bodies of text (yet)

i've had intermittent failures typing small chunks of text (email addresses) :/

matthewmueller:
I'll be adding this once there's a version bump in electron: https://github.com/atom/electron/pull/4080

:-)

stephenmathieson:
rad dude =)

the-t-in-rtf:
Glad it's not just me.  Rather than just a +1, I thought I'd share a workaround.

I often experience problems when using `type` back to back, where part of the first field's text is intermittently typed into the second field.  For example:

```
nightmare.type(".username", "username").type(".password", "password");
```

[...]

PR #415, opened 2015-12-26
Add yet another method to populate a text into an element

nodaguti:
As I mentioned in #414, `type()` has an issue that `done()` is called before Electron actually fires an input event.
This is an limitation of Electron's API, but it causes populating a long text into an element to fail in a certain situation.

So this pull request add a `populate` action to provide a more reliable way to populate a text.

### DIFF
--- a/lib/actions.js
+++ b/lib/actions.js
@@ -153,6 +153,22 @@ exports.type = function() {
 };
 
 /**
+ * Populate a text into an element
+ *
+ * @param {String} selector
+ * @param {String} text
+ * @param {Function} done
+ */
+
+exports.populate = function(selector, text, done) {
+  debug('populate() %s into %s', text, selector);
+  this.evaluate_now(function(selector, text) {
+    var element = document.querySelector(selector);
+    element.value = text;
+  }, done, selector, text);
+};
+
+/**
  * Check a checkbox, fire change event
  *
  * @param {String} selector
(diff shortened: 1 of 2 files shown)


matthewmueller:
yikes, i had a feeling that API is asynchronous. i don't like adding two methods to do nearly the same thing, but we're definitely going to need to solve this. i'll open an issue on electron to see if they have any advice.

lib/actions.js
```diff
@@ -153,6 +153,22 @@ exports.type = function() {
...
+ * @param {String} selector
+ * @param {String} text
+ * @param {Function} done
+ */
+
+exports.populate = function(selector, text, done) {
+  debug('populate() %s into %s', text, selector);
+  this.evaluate_now(function(selector, text) {
+    var element = document.querySelector(selector);
+    element.value = text;
```
yoz: In 2.0.9 (I think) the old version of type() also called element.focus() before and element.blur() after setting the value here.

yoz:
As reported in #394, the new version of type() breaks backward compatibility, so for those applications where it's preferable to simply set the value, instead of sending keystrokes and potentially appending to an existing value, it would be good to have a supported method that works the way type() used to.

Workarounds of course include writing, rather verbosely,
`nightmare.evaluate(function(selector, value) { document.querySelector(selector).value = value; }, selector, value);`
(neglecting the focus/blur that I mentioned above)

matthewmueller:
thanks for this PR, i've added an `.insert()` command based on the feedback here.

.type() should be more robust now, but i recommend you use `.insert()` if you don't care about the keyboard events.

PR #420, opened 2015-12-29
page event listener fix

antpaw:
untested

### DIFF
--- a/lib/nightmare.js
+++ b/lib/nightmare.js
@@ -108,8 +108,8 @@ function Nightmare(options) {
     process.exit(1)
   })
 
-  this.child.on('page', function(type, arg1, arg2) {
-    log('page-' + type, arg1, arg2);
+  this.child.on('page', function(type) {
+    log.apply(null, ['page-' + type].concat(sliced(arguments, 1)));
   });
 
   // proporate events through to debugging
--- a/lib/nightmare.js
+++ b/lib/nightmare.js
@@ -108,12 +108,8 @@ function Nightmare(options) {
     process.exit(1)
   })
 
-  this.child.on('page-error', function(errorMessage, errorStack) {
-    log('page-error', errorMessage, errorStack);
-  });
-
-  this.child.on('page-alert', function(message){
-    log('page-alert', message);
+  this.child.on('page', function(type, arg1, arg2) {
+    log('page-' + type, arg1, arg2);
   });
 
   // proporate events through to debugging


matthewmueller:
what does this fix?

antpaw:
it logs to the bash the page events like alert and error with `DEBUG=nightmare:log`

rosshinkley:
The change to `page` from `page-error`, `page-log` etc was an intentional one.  Take a look at the default preload script, specifically at how [page errors are handled](https://github.com/segmentio/nightmare/blob/master/lib/preload.js#L6-L8) and consequently [transmitted to Nightmare](https://github.com/segmentio/nightmare/blob/master/lib/runner.js#L90-L92).  All of the messages are (to the best of my knowledge) still being logged to `debug`, but are logically grouped together using the same event.

antpaw:
it was intentional and i understand this change how ever only the child process does this correct (as you mentioned). https://github.com/segmentio/nightmare/blob/master/lib/runner.js#L90-L92 

the parent process doesn't listen to `page` it listens to `page-error` and thats what die pr is fixing. https://github.com/segmentio/nightmare/blob/master/lib/nightmare.js#L111-L117

rosshinkley:
@antpaw Ah, I see.  I must have misread the original diff.  It looked like in passing it was adding the `page-*` events back.  At any rate, I think you are correct. The parent process should listen to `page`.  +1. :)

One nit: I think the page handler should handle variable arguments.  Something like:

```js
this.child.on('page', function(){                                                               
  log.apply(null, ['page'].concat(sliced(arguments)))                                           
}); 
```
... or if the `page-*` convention should be kept for `debug`, something like:

```js
this.child.on('page', function(){                                                               
  log.apply(null, ['page-'+arguments[0]].concat(sliced(arguments,1)))                           
});
```

Re testing, the tests for ensuring `page` is handled are already there ([alert](https://github.com/segmentio/nightmare/blob/master/test/index.js#L688-L694), [prompt](https://github.com/segmentio/nightmare/blob/master/test/index.js#L704-L721), and [confirm](https://github.com/segmentio/nightmare/blob/master/test/index.js#L704-L721) are good examples), but simply test that the events are emitted from the child and not submitted to debug.  Testing debug output would probably require using a mock (like mockery).

matthewmueller:
> I think the page handler should handle variable arguments

+1

then i think we're good to go :-)

matthewmueller:
thanks!

Issue #391, opened 2015-12-17
Support a way to extend the electron process


matthewmueller:
initial thoughts:

- support adding requires that pass in the window

```js
electron.plugins.forEach(function (plugin) {
  require(plugin)(win)
})
```

This would happen after the browser window has been initialized

rosshinkley:
It might also make sense to pass `parent`.  That would grant implementing custom actions in Nightmare with custom downstream actions in the Electron process.  Thoughts?

---

Pondering out loud: is there a reason to prefer using `require`?  If so, why does `action` not offer a `require` option?  Would the plugin paths be passed in with options, or added later?

bjrmatos:
It would make more sense to implement #393 and let nigthmare focus on its main job

bjrmatos:
inevitably, people will start to ask more and more features outside nightmare's job, it will be more easy to let nigthmare run as a electron module and developers can do what they want with the electron process

matthewmueller:
@rosshinkley i don't see a way you'd be able to pass `parent`, it's transferring over IPC and parent isn't serializable. 

The reason for doing the require is that we could basically tell the electron process to require a file on it's own. I don't like the solution all that much, but I haven't come up with a better one.

rosshinkley:
@matthewmueller did I misunderstand?  I thought the intent was passing paths to require over IPC (or possibly, as command-line arguments akin to `paths`) and requiring them in the electron process, not vice-versa.  No IPC serialization required.  Parent should then be passable.

I'll need to take another pass at this when I'm fresh and have coffee.  I'll try to carve out some time.

matthewmueller:
>  I thought the intent was passing paths to require over IPC (or possibly, as command-line arguments akin to paths) and requiring them in the electron process, not vice-versa. No IPC serialization required. Parent should then be passable.

Ahh, sorry I thought you meant `parent` as in the parent process instance. Yep the original intent is passing paths to files and requiring them in the electron process, though I don't necessarily like this approach, so if you think of something better, I'm all ears. Can you clarify what you mean by `parent` because I'm still not fully sure?

> I'll need to take another pass at this when I'm fresh and have coffee. I'll try to carve out some time.

Haha yah me too, answering issues at this hour is ill-advised :-P

rosshinkley:
@matthewmueller I should have been more specific, and the naming conventions are also probably not helping: in `runner`, `parent` is defined as the IPC process.

[...]

Issue #379, opened 2015-12-10
Provide custom parameters for actions


antpaw:
The new actions feature is great, but we need to be able to pass in arguments to the callback
so we can do something like this:

````js
Nightmare.action('evaluValue', function*(done, value, cb) {
  var test = yield this.evaluate(cb);
  assert.strictEqual(test, value, 'evaluValue() ' + errorMessageHelper(value, cb));
  done();
});
````

**Comment by matthewmueller** [comment]
hm, did you try:

```js
Nightmare.action('evaluValue', function*(value, cb, done) {
  var test = yield this.evaluate(cb);
  assert.strictEqual(test, value, 'evaluValue() ' + errorMessageHelper(value, cb));
  done();
});

Nightmare().evaluValue('some value', function cb() { ... } ) 
```

?

antpaw:
no sorry didnt tried that, though done is the first arg, this works indeed. can you help me out with this code here? thanks

```js
var url = 'http://google.com';
Nightmare.action('googleto', function(page, done) {
  this.goto(url + page);
  return done();
});

var t = yield Nightmare().googleto('/').evaluate(function(){
  document.querySelector('a').click();
}).title()
console.log(t);

**Comment by matthewmueller** [comment]
hmm, i'm not sure. that's just not working at all?

**Comment by antpaw** [comment]
this example just hangs (white screen in electron) in the air an doesn't go to the google page, the callback function of vo is never called

    var Nightmare = require('nightmare');
    var vo = require('vo');

    vo(function* () {
      var url = 'http://google.com';
      Nightmare.action('googleto', function(page, done) {
        this.goto(url + page);
        return done();
      });

      return yield Nightmare({show: true}).googleto('/').title();
    })(function (err, result) {
      if (err) return console.log(err);
      console.log(result);
    });



this code works as expected and writes 'Google' to the bash

    var Nightmare = require('nightmare');
    var vo = require('vo');

    vo(function* () {
      var url = 'http://google.com';
      return yield Nightmare({show: true}).goto(url).title();
    })(function (err, result) {
      if (err) return console.log(err);
      console.log(result);
    });

**Comment by rosshinkley** [comment]
The `DEBUG` output from above yields a clue:

```
nightmare queueing action "googleto" +0ms
nightmare queueing action "title" +5ms
nightmare running +3ms
nightmare queueing action "goto" for http://google.com/ +607ms
nightmare:actions .title() getting it +4ms
```

[...]

PR #425, opened 2016-01-04
proposal: extending electron

rosshinkley:
This is a possible implementation for solving #391.  Not _exactly_ what was requested, but hopefully close.

Some things to ponder, in no particular order:

1. This amends the instance, not the prototype.  I did this largely because Nightmare does not share Electron instances between Nightmare instances.  It didn't seem sensible to force the prototype to have Nightmare actions for which there were no Electron actions.  This also allows separate instances of Nightmare to use different plugins.
1. Adding Nightmare and Electron actions together makes more sense to me than adding Electron actions alone.  Thoughts?
1. I readily acknowledge this sort-of-kind-of steps on `.action()` (and to a lesser extent, `.use()`).  I considered updating `.use()` for plugins, but decided not to introduce breaking changes for this proposal.  Suggestions for naming conventions or restructuring welcome.
1. Constructed function scoping works a little differently.  This is why the constructed function binds to an object that passes in `require` (it's _slightly_ cleaner than using `eval`).  This allows `require` calls to be made in the Electron action.  I considered stuffing that object with `parent`, `win`, etc - this would make the Electron actions look closer to what is in `runner.js`, but decided to leave them as explicit parameters for the time being.  Thoughts?


matthewmueller:
@rosshinkley thanks for putting this together! I really have to think about this one a bit. After I get back from holiday vacation, I will take a closer look. My initial thoughts: 

> This amends the instance, not the prototype. I did this largely because Nightmare does not share Electron instances between Nightmare instances. It didn't seem sensible to force the prototype to have Nightmare actions for which there were no Electron actions. This also allows separate instances of Nightmare to use different plugins.

This could be confusing with how `.action` works, but is sort of inline with `.use`. I'm not crazy about having all these sorta different, but obscure ways to extend nightmare. I think it was discussed a bit in https://github.com/segmentio/nightmare/issues/379. 

I'm not sure the best way to fix this atm but I think where we're headed is going to lead to more confusion. It could just be a documentation thing though.

> Adding Nightmare and Electron actions together makes more sense to me than adding Electron actions alone. Thoughts?

Yah, I think you're right, but maybe actions/use could be nightmare-side only and there's an optional electron side.

> I readily acknowledge this sort-of-kind-of steps on .action() (and to a lesser extent, .use()). I considered updating .use() for plugins, but decided not to introduce breaking changes for this proposal. Suggestions for naming conventions or restructuring welcome.

Need to think on this more. But in general, I agree. It needs to be consolidated a bit more.

> Constructed function scoping works a little differently. This is why the constructed function binds to an object that passes in require (it's slightly cleaner than using eval). This allows require calls to be made in the Electron action. I considered stuffing that object with parent, win, etc - this would make the Electron actions look closer to what is in runner.js, but decided to leave them as explicit parameters for the time being. Thoughts?

Explicit params is probably fine if there's not going to need to be any additional params (or different ordering). If not, we should probably just pass a params object in.

matthewmueller:
Some general aestetic thoughts: I'm not crazy about passing an object in with `nightmareAction`, `electronAction`. I think two functions would make more sense since folks are used to `.evaluate(browser_fn, server_fn)` but I'd be open to other solutions too :-)

Thanks again for getting this started though. I think this is going to be a great addition.

rosshinkley:
@matthewmueller no problem, happy (at least to try) to help.

>This could be confusing with how .action works, but is sort of inline with .use. I'm not crazy about having all these sorta different, but obscure ways to extend nightmare. I think it was discussed a bit in #379.

I thought something eerily similar, and I actually had a point about that originally that got edited out as I couldn't make it constructive enough.  I opted to go the way I did just to get the gist of the idea on virtual paper and not stomp on stuff that already existed.  I agree that, as it stands, this will be _very_ confusing, even with more/better documentation.

>Need to think on this more. But in general, I agree. It needs to be consolidated a bit more.

I think I'm going to go ahead and update this to put all of this into `action` - that seems to be a logical consolidation step.  I also think it'll help to have a rough cut of what that would look like.  More on that when I finish.

>Explicit params is probably fine if there's not going to need to be any additional params (or different ordering). If not, we should probably just pass a params object in.

I don't _think_ there will be additional parameters, but the required signature is a bit verbose.  I'm also not crazy about wrapping the function in a context - it makes the parameters feel a bit "magic", but I thought I'd put it on the table as an option.  

>Some general aestetic thoughts: I'm not crazy about passing an object in with nightmareAction, electronAction.

A parameters hash is a bit... icky?  I'm not enthusiastic about it either.  As an aside, the reason I did it that way with `plugin` was a personal preference: I'll use a hash when I don't know what the final signature will be.  It's a vestige of experimentation. :)

rosshinkley:
@matthewmueller This update keeps the additions of the extensions together, as well as consolidates `.action()` and `.plugin()` into a single `.action()` function.  I've also kept the order of parameters similar to `.evaluate()` where the (optional) client method comes before the Nightmare action.

There are a couple of differences (and possible pitfalls?) in this update.  For starters, to keep the current behavior of `.action()` consistent, I decided it made the most sense to have the Electron plugin(s) be added for all instances.  The child actions are cached on the nightmare constructor, similar to how namespacing for `.action()` works.  Consequently, the addition of those child actions are added to the queue in the constructor such that all child action additions are executed prior to the first user-specified action.  (And they're silent, something I've got mixed feelings about.)

Thoughts?

matthewmueller:
aestetically and complexity-wise, I like this **a lot** more :-). can you explain what you see as a pitfall of adding the electron plugin to all instances?

rosshinkley:
@matthewmueller When I wrote that sentence about possible pitfalls, I had two things in mind:

1. I was changing how `.plugin()` behaved considerably - originally, you could add Electron functionality piecemeal.  This meant the execution time/complexity to add Electron extensions was only as big as the instance required, whereas now, actions for _every_ plugin are queued for every instance before execution can even start.  On reflection, since the plugin action is uniform (except in what's actually transferred), I'd bet the performance impact is (almost) negligible.  At any rate, it's not anywhere near dealbreaking, only the fringiest (is that a word?) of fringe cases would be using multiple Nightmare instances with different bolt-ons.  TLDR: adding plugins to every instance isn't going to be a problem. :)
2.  I know I mentioned this already: queued plugin actions execute silently before the first user call.  Should they at least make a `debug` call or something?

Finally: anything else I can do to help this PR along?

rosshinkley:
@matthewmueller I went ahead and added the debug, and also went ahead and merged in the recent changes.  Special thanks to @emars for fixing the build, something I just hadn't had time to do. :)

rosshinkley:
@matthewmueller I've gone ahead and thrown together a couple of proposed plugins: [additional events](https://github.com/rosshinkley/nightmare-custom-event) (from #376) and [download management](https://github.com/rosshinkley/nightmare-download-manager) (from #382).  From that, I got to use this PR for "real" applications, and as such, made a couple of touchups to include options in the Electron method, as well as attaching the options hash to the instance.  This allows for making/expecting options in the Nightmare options for plugin use.

rosshinkley:
A note on `require`'s context: The Electron context does have `require` attached, but it's worth noting that it'll require things from the local Nightmare install's node modules, not the plugin's node modules due to how the Electron function is created.  Plugin-installed modules can be required if they are specified with absolute or relative paths passed from the Nightmare plugin to the Electron plugin.

An example may be helpful.  Say we wanted to do some date manipulation in Electron and wanted to use Moment to help:
```js

Nightmare.action('requireSomething',
    parent.on('require-example', function(){
      parent.emit('log', 'trying to require moment');
      var moment = require('moment');
      parent.emit('log', moment().format('YYYY-MM-DD'));
      parent.emit('require-example');
    })
    done();
    return this;
  },
  function(done) {
    this.child.once('require-example', done);
    this.child.emit('require-example');
  });
```
... won't work.  You'll get an exception from `runner` saying it couldn't find moment.  However, if you pass the module path down:

```js
var Nightmare = require('nightmare'),
    path = require('path');

Nightmare.action('requireSomething',
    parent.on('require-example', function(modulePath){
      var path = require('path');
      parent.emit('log', 'trying to require moment');
      var moment = require(path.resolve(modulePath, 'moment'));
      parent.emit('log', moment().format('YYYY-MM-DD'));
      parent.emit('require-example');
    })
    done();
    return this;
  },
  function(done) {
    this.child.once('require-example', done);
    this.child.emit('require-example', path.resolve(__dirname, 'node_modules'));
  });
```
... allows Moment to be required.

---

I considered trying to make `require` automagically work: the `.action()` method would need to determine the caller location (which can be done by sifting through an artificial stacktrace, or use of `module.parent`, both of which have problems).  Then, Electron's `require` would need to be wrapped in a function that determines if it's a bare package name or path, require, and return the package, and I'm sure there's other complexity buried there.  

In other words, it's complicated, and probably too much complexity to introduce when 1) using `require` from Electron is probably going to be rare and 2) passing the path explicitly avoids complication entirely.

I'm certainly open to suggestions or thoughts. :)

---

Is it worth including something akin to the above in the documentation to clear up possible problems with `require` from Electron?

PR #444, opened 2016-01-20
Update use of electron APIs to kill deprecation warnings for 1.0

inversion:
See http://blog.atom.io/2015/11/17/electron-api-changes.html


matthewmueller:
thanks!

PR #445, opened 2016-01-20
If a reject function is not passed to .then, throw the error

inversion:


matthewmueller:
hmm, can you give me an example of when you'd want to throw? these are all asynchronous events, so i don't think throwing would actually work as expected.

inversion:
This is because there is no default reject function, so if you fail to pass one then you will get an unhelpful error à la `undefined is not a function`. Throwing it allows your `uncaughtException` handler to log out the actual error.

Mr0grog:
@inversion I think the issue you were trying to solve here was fixed when @matthewmueller changed `.then()` to return a promise: https://github.com/segmentio/nightmare/commit/54f81d3dd9e6e4dad861f749a4a3064e2ce98dd0

If so, it would probably help to close this PR.

inversion:
With that change it will now log out properly so closing.

```
new Promise(function(fulfil, reject) { reject(new Error('foo')); }).catch(undefined);
// Promise {[[PromiseStatus]]: "pending", [[PromiseValue]]: undefined}
// Uncaught (in promise) Error: foo(…)
```

PR #447, opened 2016-01-21
Use template strings for javascript templates

inversion:


rosshinkley:
@inversion I'm definitely in favor of the change, but it needs to be updated to be current.  Would you mind updating this?

inversion:
@rosshinkley now updated

rosshinkley:
Thanks, @inversion!

Issue #224, opened 2015-09-16
No output from page parsing, but no error messages


richard5mith:
I've taken the cnn.js example, and my only change is changing the require line from ../nightmare to nightmare, but when I run it I only get...

> DEBUG=nightmare node --harmony cnn.js

    nightmare queueing action "goto" +0ms    
    nightmare queueing action "evaluate" +5ms

I don't ever seem to get the title of the page.

I've tried various other examples from Nightmare < 2 from around the web, and I can't get any of them to work either. Nothing ever errors, I just never seem to get any screenshot, pdf or elements pulled off the page in the output.

After 4 hours of bashing my head, I'm not sure what else to try.

idrakimuhamad:
I tried this and got it returned properly.

    nightmare queueing action "goto" +0ms
    nightmare queueing action "evaluate" +1ms
    Breaking News, U.S., World, Weather, Entertainment & Video News - CNN.com

But if I tried with a website that have a SLL problem, like below, it hang there for, forever. 

        vo(run)(function(err, result) {
          if (err) throw err;
        });

        function *run() {
            var title = yield nightmare
            .goto('https://cas2.northport.com.my/myapp/wa/r/nmbctr')
            .evaluate(function() {
                return document.title;
            });
            console.log(title);
            yield nightmare.end();
        }

It just hang here. No error, and electron no exited.

        nightmare queueing action "goto" +0ms
        nightmare queueing action "evaluate" +1ms

richard5mith:
What's your environment for running the cnn.js script?

Entirely headless server, or in a terminal in an existing X session, or OS X etc?

pastawoua:
I do have the same issue on a headless server. Everything runs correctly on OSX but not on my build server.

matthewmueller:
Yikes, major issue. 

![img](https://cldup.com/dAy_dE48-X.png)

Seems like the electron linux binding is broken or something. Anyone have any ideas?

richard5mith:
Odd that at least you get an error message when I don't. Mine just fails silently.

My guess is that it can't get the framebuffer, which is where xvfb would normally come in. But I've tried running electron with that too and just get gtk errors.

matthewmueller:
Okay. I'm not exactly sure how to codify this right now (or what layer to support this in), but to get it working on the official node (docker) image, you need to do the following:

```
# Install dependencies
apt-get update &&\
    apt-get install -y libgtk2.0-0 libgconf-2-4 \
[...]

PR #459, opened 2016-01-30
add debug to inspect how electron exited

maxbeatty:
Yesterday, I ran up against #224, and before finding all of the helpful info there (_thank you everyone_), I was doing my own debugging based on the [exit code](http://tldp.org/LDP/abs/html/exitcodes.html) of the electron child process. This should hopefully make it easier to identify potential issues when ["it just quits."](https://github.com/segmentio/nightmare/issues/224#issuecomment-177268035)

If you're seeing exit code 127, you may not have electron installed correctly.
If you're seeing exit code 126, you may not have all the necessary dependencies for electron (:wave: gtk).
If you're seeing exit code 1, xvfb is a likely suspect in my experience.
If you're seeing exit code 0, electron appears to be working fine and it's something else.

Do you agree the exit code is helpful when debugging? Should these possible problems be included in the source, README, or wiki?

### DIFF
--- a/lib/nightmare.js
+++ b/lib/nightmare.js
@@ -69,8 +69,15 @@ function Nightmare(options) {
   });
 
   this.proc.on('close', function(code) {
-    debug('electron child process exited with code ' + code)
-  })
+    var help = {
+      127: 'command not found - you may not have electron installed correctly',
+      126: 'permission problem or command is not an executable - you may not have all the necessary dependencies for electron',
+      1: 'general error - you may need xvfb',
+      0: 'success!'
+    };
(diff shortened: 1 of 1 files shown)


matthewmueller:
Yes, I love this idea! But I think it should be baked into the code itself instead of in the readme. Basically just: 

```js
var errors = { ... }
...
debug(errors[code])
```

maxbeatty:
:+1: I like when software tries to help me directly, too

maxbeatty:
Is there anything else I can do for this PR before merging?

PR #464, opened 2016-02-05
Generic MouseEvent trigger function same #84

Max93:
This pull is like #84 
I creating another pull request because i deleted the oldest one(don't ask me why please).
Past the response from @reinpk 

"thanks @Max93! sorry i never got to this with v1... it looks like an interesting abstraction for other events... what events were you interested in? closing for now but may bring the idea into v2 with more info here. thanks!"

Thanks again @reinpk

### DIFF
--- a/lib/actions.js
+++ b/lib/actions.js
@@ -74,15 +74,27 @@ exports.exists = function(selector, done) {
  */
 
 exports.click = function(selector, done) {
-  debug('.click() on ' + selector);
-  this.evaluate_now(function (selector) {
-    document.activeElement.blur();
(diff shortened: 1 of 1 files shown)


matthewmueller:
cool i like it :-)

what kind of events are available by making this more generic? I can think of click and dblclick. would this also include mousedown, mousemove, etc?

additionally, maybe we can rename it to `.mouse(selector, event)`?

reinpk:
def open to this abstraction, would be great to see something like `doubleclick` or similar implemented to make the abstraction worthwhile though. maybe re-open with a PR that adds something like that?

PR #465, opened 2016-02-06
Fixes multiple instances of nightmare using the same waitTimeout value

Crazometer:
Previously if you had multiple instances of nightmare spun up and all using .wait() they would all share a single waitTimeout counter. This resulted in some elements firing incorrectly and others not firing when they should. Instead the waitTimeout value has been attached to the nightmare instance which also makes it observable. Alternatively you could create a closure but I think this is nicer.

### DIFF
--- a/lib/actions.js
+++ b/lib/actions.js
@@ -356,28 +356,30 @@ function waitelem (self, selector, done) {
  */
 
 var _timeoutMs = 250;
-function waitfn (self, fn/**, arg1, arg2..., done**/) {
-  self._waitMsPassed = self._waitMsPassed || 0;
-  var args = sliced(arguments);
-  var done = args[args.length-1];
-  var waitDone = function (err, result) {
-    if (result) {
-      self._waitMsPassed = 0;
-      return done();
-    }
-    else if (self.optionWaitTimeout && self._waitMsPassed > self.optionWaitTimeout) {
-      self._waitMsPassed = 0;
-      return done(new Error('.wait() timed out after '+self.optionWaitTimeout+'msec'));
-    }
-    else {
-      self._waitMsPassed += _timeoutMs;
-      setTimeout(function () {
-        waitfn.apply(self, args);
-      }, _timeoutMs);
-    }
-  };
-  var newArgs = [fn, waitDone].concat(args.slice(2,-1));
-  self.evaluate_now.apply(self, newArgs);
+function waitfn() {
+  var waitMsPassed = 0;
+  return tick.apply(this, arguments)
+
+  function tick (self, fn/**, arg1, arg2..., done**/) {
+    var args = sliced(arguments);
+    var done = args[args.length-1];
+    var waitDone = function (err, result) {
+      if (result) {
+        return done();
+      }
+      else if (self.optionWaitTimeout && waitMsPassed > self.optionWaitTimeout) {
+        return done(new Error('.wait() timed out after '+self.optionWaitTimeout+'msec'));
+      }
+      else {
+        waitMsPassed += _timeoutMs;
+        setTimeout(function () {
+          tick.apply(self, args);
+        }, _timeoutMs);
+      }
+    };
+    var newArgs = [fn, waitDone].concat(args.slice(2,-1));
+    self.evaluate_now.apply(self, newArgs);
+  }
 }
 
 /**
--- a/lib/actions.js
+++ b/lib/actions.js
@@ -355,22 +355,22 @@ function waitelem (self, selector, done) {
  * @param {Function} done
  */
 
-var _waitMsPassed = 0;
 var _timeoutMs = 250;
 function waitfn (self, fn/**, arg1, arg2..., done**/) {
+  self._waitMsPassed = self._waitMsPassed || 0;
   var args = sliced(arguments);
   var done = args[args.length-1];
   var waitDone = function (err, result) {
     if (result) {
-      _waitMsPassed = 0;
+      self._waitMsPassed = 0;
       return done();
     }
-    else if (self.optionWaitTimeout && _waitMsPassed > self.optionWaitTimeout) {
-      _waitMsPassed = 0;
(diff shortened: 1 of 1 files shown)


Mr0grog:
*(Apologies for just poking my head in here, but for some reason this piqued my attention…)*

FWIW, I think this should be a closure—not because it shouldn't be observable, but because it's not instance-level information and could result in conflicts. If you don't use the `yield`-based API, you can have two chains of actions going at once and thus two different `waitfn` functions running at the same time. You wouldn't want the timeout from one to trample the other:

```js
var page = Nightmare();

// separate out the page load so the tests run while the page is loading
page('http://yahoo.com');

// these two tests actually run at the same time and race for completion.
page
  .wait(function() { return document.activeElement.id === "uh-search-box" })
  .then(function() {
    console.log("Search box is focused");
  });
page
  .wait(function() { return document.querySelector(".darla").firstChild.nodeName === "IFRAME" })
  .run(function() {
    console.log("First ad is loaded");
  });
```

Granted, the current functionality already shares timers in a way that causes the same problem, but it seems like it would make sense to fix this at the same time.

Crazometer:
@Mr0grog, thanks for poking your head in here. I can see the use case for the point above but I'm kind of confused on the implementation. Aren't the `.wait()`'s placed in a queue?

As a sidenote, when I was testing things I found that setting `var _timeoutMs = 1;` could cause strange results. Noteably both would fire in the beneath
```
page
  .wait(function() { return document.querySelector(".darla").firstChild.nodeName === "IFRAME" })
  .run(function(err, res) {
    console.log("First ad is loaded");
  });

page
  .wait(function() { return document.activeElement.id === "notAnId" })
  .run(function(err, res) {
    console.log("Search box is focused");
});
```

Mr0grog:
> Aren't the .wait()'s placed in a queue?

Yes, but every time you call `run()` or `then()`, the current queue is split off and run independently. So:

1. Queue up some operations
2. Call `run()`
    1. Inside run, a copy of the the nightmare instance's queue is made
    2. The nightmare instances queue is reset to an empty list
    3. The copy of the queue starts getting worked on—all the queue work is entirely inside a closure in `run()`.
3. Queue up some more operations
    - These are placed on the freshly empty queue made in (2.ii), so they aren't really queued up behind the first set of operations
4. Call `run()` again
    - Now we do the same steps as in (2). As a result this set of operations is running simultaneously with (instead of after) the operations from step (1). It's all about when you call `run()`.

> As a sidenote, when I was testing things I found that setting var _timeoutMs = 1; could cause strange results. Noteably both would fire in the beneath

Hmmm, I'm not entirely sure what you mean by this. You should always have both of those `run()` callbacks fire, regardless of timeout (unless `_timeoutMs` was infinity and one of the wait functions never returned `true`).

I may have worded things poorly when I said "race for completion" earlier—they both run to completion, just simultaneously instead of in sequence, so they could finish in a different order than you start them (e.g. you could get the message that the search box was focused before the one that the ad was loaded).

Crazometer:
Thanks a ton for the explanation, that really clears things up. I'll swap it over to a closure once I get the chance.

As for the `timeoutMs` thing, if you look closely at the selector in the second call you will see that it is bogus. `return false` works equally well to demonstrate the point.

Mr0grog:
d'oh! I didn't notice the test condition at all :P

Either way, both should still fire; the bad one will just do so by timing out. The `err` param should be a timeout error. Is that what you mean? Or is something else odd happening?

Crazometer:
Sorry , I should've been more specific. The exactly code I am running with `_timeoutMs` is

```javascript
var page = Nightmare({show:true});

// separate out the page load so the tests run while the page is loading
page.goto('http://yahoo.com');

// these two tests actually run at the same time and race for completion.
page
  .wait(function() { return false })
  .run(function(err, res) {
  	console.log(err);
  	console.log(res);
    console.log("Search box is focused");
})

page.wait(function() { return document.querySelector(".darla").firstChild.nodeName === "IFRAME" })
  .run(function(err, res) {
  	console.log(err);
  	console.log(res);
    console.log("First ad is loaded");

//output
//undefined
//undefined
//First ad is loaded
//undefined
//undefined
//Search box is focused
});
````

Mr0grog:
Oh, this is actually a neat little race condition/bug. The first thing you have to know for this one is that, if you ask Electron to execute JS code before the browser window has even started to load a page, it waits until the page is available to call it (looks like after DOM ready, but not 100% sure here, the implementation is in Chromium, not Electron). So this is what's happening (it's kind of complicated):

1. `page.goto('http://yahoo.com');` doesn't do anything (yet) because `run()` isn't called immediately after.
2. We add the never-will-succeed test to the queue (so now the queue is: [load page, run test]
3. We start that queue running
4. We create a new queue with the test for the ad and start *it* running.
5. Because queue 2 isn't waiting for anything, it starts *immediately,* before any page has even started to load (a browser request has been made, but the response hasn't started to be received yet).
    1. Nightmare registers listeners for the results of the wait function (the one for the ad)
    2. Electron waits a while before executing the actual javascript. Meanwhile…
6. In queue 1, the page finishes loading and we attempt to execute our will-always-fail wait function
    1. Nightmare registers listeners for the results of the wait function—critically *these listeners do not distinguish between **which** wait function*
    2. Electron queues up the wait function to run
7. Electron knows the page is ready now, so it starts running any queued-up JS functions (at this point, that's *both* of our wait functions)
8. The first wait function runs
9. *Both* sets of listeners for the wait functions hear the results of the first wait function as if it were the wait function they were intended for. The ones for the function registered first (the one written *second* in our code) is right. The other one is wrong and reports back the wrong results.

  In your code above, that means the listeners that were set up for the will-always-fail function are actually reporting the results of the test for the ad.

That's why your results are different if you change the order. The first function you write is the second one to get registered, which means it reports the results of the first one to get registered. You'll notice different results if you call `run()` immediately after `page.goto(...)`, because then that means that the first wait function *does* get registered before the second wait function!

Crazometer:
Swapped it over to a closure. Should be good to merge now.

PR #476, opened 2016-02-14
Fix inject an event for console.log

koba04:
Currently, `on('console', function() {})` cannot catch `console.log` event when it is called in `nightmare.click(selector)`.

I guess nightmare injects wrong `ipc.send` for `console.log` at `lib/javascript.js`.
nightmare injects an event for `console.log` at `lib/preload.js` already. So the injecting at `lib/javascript.js`  seems to be unnecessary.
Additionally, it doesn't match the interface in between `lib/preload.js` and `lib/javascript.js`.

https://github.com/segmentio/nightmare/blob/master/lib/preload.js#L14

### DIFF
--- a/lib/javascript.js
+++ b/lib/javascript.js
@@ -12,19 +12,13 @@ var minstache = require('minstache');
 
 var execute = [
   "(function javascript () {",
-  "  var log = console.log;",
   "  var ipc = __nightmare.ipc;",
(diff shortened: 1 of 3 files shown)


rosshinkley:
+1.  This corrects bugs introduced in 2.1.0.

koba04:
Thanks!

Issue #468, opened 2016-02-09
Screenshots sometimes out-of-sync with actual state of web contents


Mr0grog:
In trying to use Nightmare for doing some screen captures of highly dynamic web pages, I've discovered that the screenshots are often out-of-sync with the state of the web page being captured. After a lot of investigation, it turns out that this is actually an underlying issue in Electron—it generates a screenshot by simply copying from Chrome's backing store, which is not actually guaranteed to be up-to-date (see [Electron issue 1049](https://github.com/atom/electron/issues/1049) and [this Chromium-dev discussion](https://groups.google.com/a/chromium.org/forum/#!topic/chromium-dev/o2aUtFAIlyQ)).

I believe this is probably the underlying issue behind a few other nightmare bugs, such as: https://github.com/segmentio/nightmare/issues/33, https://github.com/segmentio/nightmare/issues/129, the long wait @rosshinkley suggests adding in his solution to https://github.com/segmentio/nightmare/issues/328

But! It turns out there is a way to address this. In Electron's `BrowserWindow.webContents`, you can call `beginFrameSubscription(callback)`, which seems (in my testing) to guarantee that the backing store is up-to-date when your callback is called. So if you changed [line 252 of `lib/runner.js`](https://github.com/segmentio/nightmare/blob/master/lib/runner.js#L252) to:

```js
win.webContents.beginFrameSubscription(function(buffer) {
  win.webContents.endFrameSubscription();
  win.capturePage.apply(win, args);
});
```

You ought to have a much more deterministic result when calling `.screenshot()`.

Unfortunately, I don't have a great reduced test case for this, but I'm more than happy to submit a patch if desired.

matthewmueller:
awesome! we'd love to see a patch for a more deterministic result!

rosshinkley:
@Mr0grog Glad to finally have an answer to that question, and good eye for finding that issue!  Especially good to know it's not a framebuffer-dependent issue.

I agree: anything to make the behavior deterministic.  I do have an edge-case question, though - in my read of the documentation, the `beginFrameSubscription` callback is called for any display event.  If the screenshot is taken well after the page is rendered, will the callback still be called?

Mr0grog:
@matthewmueller Great, I'll try and put together a PR this week.

@rosshinkley That is a really good question! I will poke around and see what I can see. In the worst case, twiddling with something on the page, such as adding and removing an element, ought to work to trigger it.

Mr0grog:
[...]

PR #479, opened 2016-02-15
Synchronize screenshots using frame subscription API

Mr0grog:
This is meant to address #468.

The basic gist of the change is to add a `FrameManager` object that emits events whenever the browsers renders a new frame and use that to figure out when to call `capturePage()`.

Some notes:

- At current, this can trigger pretty massive memory consumption. Due to a bug in Electron (https://github.com/atom/electron/issues/4441), trying to unsubscribe from frames can cause a crash (and there's no way to know whether that will happen ahead of time and smartly wait until it's safe). That means that, once we start subscribing, we don't stop until the window is closed (i.e. until nightmare's `end()` method is called).

  For reference, a really simple nightmare session takes 20-30 MB of RAM on my machine, but once a screenshot is taken, that can boost to as high as 360 MB. (You can actually watch V8 manage memory here—the electron process will slowly grow to a few hundred megabytes and then get chopped back down to less than 50, then start growing again…)

  Anyway, there's already a patch on the way for this in Electron (https://github.com/atom/electron/pull/4451), but I don't know when it will land and ship. Once it does, the code here can be very easily adjusted to smartly unsubscribe if it's safe based on the version of Electron we're running in: https://github.com/Mr0grog/nightmare/blob/a179e2ed6e7b1d72d8cd518a9b464c0bf0a68220/lib/frame-manager.js#L11

- FrameManager is an event emitter, so it could pretty easily support some interesting other behaviors in the future, like rendering video of the page in action. It knows how to manage frame subscriptions in Electron when listeners are attached/detached.

- It's likely this would be better off in that future case as a readable stream, but managing push/pull/buffering vs. whenever Chromium decides to render a frame seemed potentially non-trivial (or at least require lots of testing and probably some nuance with finding an appropriate high-water mark), so I didn't bother with that for now.

Anyway, hope this helps and looks good to you all. It's covered by tests, too. Let me know if there's anything that should be changed.

### DIFF
--- a/lib/frame-manager.js
+++ b/lib/frame-manager.js
@@ -1,14 +1,13 @@
 const parent = require('./ipc')(process);
 const EventEmitter = require('events');
 const util = require('util');
+const semver = require('semver');
 
 const RENDER_ELEMENT_ID = '__NIGHTMARE_RENDER__';
 
-// NOTE: we cannot unsubscribe from frame notifications due to:
-//   https://github.com/atom/electron/issues/4441
-// but in a future release of Electron, we should replace this with:
-//   semver.gte(process.versions.electron, '[fixed version]')
-const CAN_UNSUBSCRIBE = false;
+// NOTE: Electron crashes when unsubscribing from frames prior to v0.36.8. For
+// more info, see https://github.com/atom/electron/issues/4441
+const CAN_UNSUBSCRIBE = semver.gte(process.versions.electron, '0.36.8');
 
 module.exports = FrameManager;
 
--- a/lib/frame-manager.js
+++ b/lib/frame-manager.js
@@ -1,4 +1,6 @@
 const parent = require('./ipc')(process);
+const EventEmitter = require('events');
+const util = require('util');
 
 const RENDER_ELEMENT_ID = '__NIGHTMARE_RENDER__';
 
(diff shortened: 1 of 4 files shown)


Mr0grog:
There is a shipping version of Electron with the fix for the crash now! I've added semver support for checking the electron version to see if it's safe to unsubscribe. Alternatively, we could just bump the minimum required version of electron-prebuilt in `package.json`.

rosshinkley:
I'd vote for simply upping the Electron version, barring other changes made internally to Electron.  Is there a reason not to?

At any rate: +1 for this PR. :)

Mr0grog:
@rosshinkley No reason not to bump the version as far as I'm aware. Figured I'd go for safety first, though, since I didn't know if there was any policy around Electron versions.

rosshinkley:
@Mr0grog I can't find it with a cursory search, but I'm almost positive the version has been upped to accommodate changes like this before.  There are plenty of issues surrounding version changes where Electron changed some sort of internal convention (preference names like `web-preferences` to `webPreferences`, for example).  I think I'd still rather manage the version bump fallout over managing version compatibility, though.

Mr0grog:
@rosshinkley Makes sense, thanks for the feedback. Bumped the version and eliminated all the `CAN_UNSUBSCRIBE` stuff in f4499d0ff5145a6082c7630dfba51c100a760d1f. :thumbsup:

Mr0grog:
Howdy! It’s been a few weeks with no official feedback, so I’m just checking in. Any idea when you might get a chance to take a look at this?

mikeys:
@Mr0grog I'm using the `FrameManager` code and from time to time I still experience problems with incomplete/blank captures.
Perhaps it has to do with the fact that we rely on the first `buffer` event, but I've noticed that sometimes I get 2 and more events (depends on the complexity of the item being rendered?)
I've noticed that adding `beginFrameSubscription(debounce(receiveFrame, frameDebounce)` produces more consistent results.

Mr0grog:
> Perhaps it has to do with the fact that we rely on the first buffer event, but I've noticed that sometimes I get 2 and more events (depends on the complexity of the item being rendered?)

Well, anything that happens in the window that causes rendering (even moving the mouse around) can cause frames to be rendered, so I don’t think it would be at all surprising to see more than one frame event. I suppose it’s possible that there could also be frames already queued up at the time of subscription that we receive when first subscribing, so you might also be seeing that.

Given that, I can definitely see how debouncing could be improve accuracy (though at a significant time cost). You should submit a PR! (FYI I’m not a maintainer/committer here; I can’t approve or merge anything for you.)

Issue #477, opened 2016-02-15
can't get the example of evaluate() to work


max-l:
I run the exact code snippet  and evaluate() returns null
(i.e. the "console.log(result);" below returns null)

I tried many other examples using vo(), and I can never obtain the result of evaluate()

nightmare 2.1.6
node 4.3.0 

```
var Nightmare = require('nightmare');
var vo = require('vo');

vo(function* () {
  var nightmare = Nightmare({ show: true });
  var link = yield nightmare
    .goto('http://yahoo.com')
    .type('input[title="Search"]', 'github nightmare')
    .click('#uh-search-button')
    .wait('.ac-21th')
    .evaluate(function () {
      return document.getElementsByClassName('ac-21th')[0].href;
    });
  yield nightmare.end();
  return link;
})(function (err, result) {
  if (err) return console.log(err);
  console.log(result);
});
```

This other example (found in issue #347) also outpuns null on the console

coult it be that something has changed in recent nightmare versions ? 

```

var Nightmare = require('nightmare'),
    vo = require('vo');

vo(function * () {

    var outputs = [];
    var inputs = ['2', '3', '4', '5', '6', '7'];

    var nightmare = new Nightmare({
            show: true
        })
        .useragent('Mozilla/5.0 (Macintosh; Intel Mac OS X 10_10_4) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/46.0.2490.80 Safari/537.36')
        .viewport(600, 770)
        .goto('http://www.adrianparr.com/temp/square.html')
        .wait()

    for (i = 0; i < inputs.length; i++) {
        outputs.push(yield nightmare
            .type('#input', inputs[i])
            .click('#submit')
            .wait(1000)
            .evaluate(function() {
                return document.getElementById("output")
                    .innerHTML;
            }));
    }

    yield nightmare.end();
    return outputs;
})(function(err, result) {
    console.log(result);
});
```

Mr0grog:
Hey @max-l, it looks like the problem you're having is actually a change in Vo, not Nightmare. In Vo version 2 (just released yesterday!), it says: "In the first version return values were passed on to the next function by default. In this version the default uses fixed arguments you pass in when you call vo."

So, you have to use the `pipeline()` method to get the v1 style behavior:

```js
var Nightmare = require('nightmare');
var vo = require('vo');

vo.pipeline(function* () {
  var nightmare = Nightmare({ show: true });
  var link = yield nightmare
[...]

PR #482, opened 2016-02-15
Fix Readme examples to work with vo 2.0

Mr0grog:
Figured the docs should be updated so more people don't have trouble as in #477.

I wound up making significant changes to the last example so that it would actually still work safely in the case of an error. It seems that, if the vo callback that is handling the error throws, then vo just continues execution as normal in the `run` function and doesn't raise the error. So `undefined` is logged and it looks like no error happened, but no title was found, either. I changed it to log the error, but that also means the handler has to call `process.exit()` otherwise it hangs because `nightmare.end()` is never called because of the error getting the title. Maybe it would just be better to use try/catch/finally instead, since generators allow that:

```js
vo(run)();

function *run() {
  var nightmare = Nightmare();
  try {
    var title = yield nightmare
      .goto('http://cnn.com')
      .evaluate(function() {
        return document.title;
      });
    console.log(title);
  }
  catch(error) {
    console.error(error);
  }
  finally {
    yield nightmare.end();
  }
}
```

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -29,7 +29,7 @@ Let's search on Yahoo:
 var Nightmare = require('nightmare');
 var vo = require('vo');
 
-vo(function* () {
+vo.pipeline(function* () {
   var nightmare = Nightmare({ show: true });
   var link = yield nightmare
     .goto('http://yahoo.com')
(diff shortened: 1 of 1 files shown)


matthewmueller:
Ahh sorry about this, I forgot that we were installing vo from latest. I went ahead and removed vo altogether from the example. We'll be recommending you use `.then()` from here on out.

Mr0grog:
So… the new first example doesn't work because the `end()` call happens simultaneously with the queue being used to get the link. It kills the electron process before the queue can complete.

The second example (down near the end under the heading "Execution") is still using vo.

Mr0grog:
Also, *both* those examples should probably:

`var Nightmare = require('nightmare')`

rather than:

`var Nightmare = require('./');` (the first one) and
`var Nightmare = require('../nightmare');` (the last one).

PR #487, opened 2016-02-16
Kill electron process when Nightmare's process exits/dies

### LINKED ISSUE #474: Kill electron instance when nightmare's process exits
Since I started working on #468, I noticed that exiting a process that is using Nightmare doesn't end the electron process that Nightmare is controlling. At first, I had no idea this was happening because I was running nightmare with `{show: false}`, but I soon discovered I had *a lot* of electron processes floating around:

lotsofelectrons

Unless there's an important reason for that, it seems like simply adding a listener to the process's `exit` event in Nightmare's constructor that kills electron would be both simple and helpful:

```js
process.on('exit', function() {
  self.proc.disconnect();
  self.proc.kill();
});
```

That is, I don't think I'm unique in sometimes forgetting to call `end()` ;)

Mr0grog:
If the process Nightmare is running in exits or is killed (via a graceful process.exit() call, interrupt signal, etc), this ensures that the underlying Electron process Nightmare is managing also gets closed and cleaned up so users don't have have tons of rogue Electrons running.

Fixes #474.

### DIFF
--- a/lib/nightmare.js
+++ b/lib/nightmare.js
@@ -74,7 +74,14 @@ function Nightmare(options) {
     endInstance(self);
   });
 
-  process.on('exit', endInstance.bind(null, self));
+  // if the process nightmare is running in dies, make sure to kill electron
+  var endSelf = endInstance.bind(null, self);
+  process.on('exit', endSelf);
+  process.on('SIGINT', endSelf);
+  process.on('SIGTERM', endSelf);
+  process.on('SIGQUIT', endSelf);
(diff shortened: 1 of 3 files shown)


rosshinkley:
+1 for the PR, and a double +1 for the remote branch name. ;)

matthewmueller:
thanks @Mr0grog !

Mr0grog:
:+1: No problem, happy to help out.

PR #496, opened 2016-02-22
Enable support for IFrames, fixes #203

### LINKED ISSUE #203: Interacting with elements inside an iframe

tiangolo:
Enable support for IFrames, fixes #203

It creates two variables in the browser object `__nightmare`: `currentDocument` and `rootDocument`.

`rootDocument` is a reference to the main `document` used normally.

`currentDocument` starts as a reference to the main `document` but is updated to be a reference to the `document` of an iframe element by the action `Nightmare.prototype.enterIframe(selector)`.

The method `Nightmare.prototype.enterIframe(selector)` also works from inside iframes, so it allows traversing an arbitrary tree of iframes.

The method `Nightmare.prototype.exitIframe()` gets out of the iframes and comes back to the root `document` element.

An example would be:

```
var nightmare = Nightmare();
nightmare.goto('http://www.w3schools.com/tags/tryit.asp?filename=tryhtml_iframe').then();

nightmare.title().then(function(res){console.log(res)});

nightmare.enterIframe('#iframeResult').enterIframe('iframe');

nightmare.title().then(function(res){console.log(res)});

nightmare.exitIframe();

nightmare.title().then(function(res){console.log(res)});

nightmare.end();
```

It would print to the console:

```
Tryit Editor v2.6
W3Schools Online Web Tutorials
Tryit Editor v2.6
```

The two additional actions (`.enterIframe(selector)` and `exitIframe()`) are also documented in the `Readme.md`.

---

The way it works is by embedding the functions called with `.evaluate_now()` in a closure (to have an independent scope that allows overriding the `document` variable) giving them a `document` variable extracted from the `__nightmare.currentDocument` in the browser. That way, other actions (methods of `Nightmare.prototype`) don't have to be modified and user's code using `.evaluate()` would also work as expected.

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -173,11 +173,12 @@ It can be called again from inside an iframe, so you can traverse a series of em
 
 To go back to the root `document`, call `.exitIframe()`.
 
-If the iframe you are entering comes from another domain, you may want to create your first `Nightmare` instance with `{'web-preferences': {'web-security': false}}` as in:
+If the iframe you are entering comes from another domain, you may want to create your first `Nightmare` instance with `{'webPreferences': {'webSecurity': false}}` as in:
 
 ```
-var nightmare = Nightmare({show: true, 'web-preferences': {'web-security': false}});
+var nightmare = Nightmare({show: true, 'webPreferences': {'webSecurity': false}});
 ```
+You may want to check those and other possible configurations in the [Electron BrowserWindow options](https://github.com/atom/electron/blob/master/docs/api/browser-window.md#new-browserwindowoptions) (the same documentation listed above in [Nightmare(options)](#nightmareoptions)).
 
 #### .exitIframe()
 Exit from all iframe elements entered with `.enterIframe(selector)` and return to the root document.
--- a/Readme.md
+++ b/Readme.md
@@ -164,6 +164,26 @@ Go forward to the next page.
 #### .refresh()
 Refresh the current page.
 
+#### .enterIframe(selector)
+Enter in the `selector` iframe element and use its `document` as the `document` for subsequent actions.
+
+Any action called after this will act as if it was called from inside the iframe element.
+
+It can be called again from inside an iframe, so you can traverse a series of embedded iframes.
+
+To go back to the root `document`, call `.exitIframe()`.
+
+If the iframe you are entering comes from another domain, you may want to create your first `Nightmare` instance with `{'web-preferences': {'web-security': false}}` as in:
+
+```
+var nightmare = Nightmare({show: true, 'web-preferences': {'web-security': false}});
+```
+
+#### .exitIframe()
+Exit from all iframe elements entered with `.enterIframe(selector)` and return to the root document.
+
+Any action called after this will act as normal.
+
 #### .click(selector)
 Clicks the `selector` element once.
 
--- a/lib/actions.js
+++ b/lib/actions.js
@@ -295,6 +295,20 @@ exports.enterToIframe = function(selector, done){
     }, done, selector);
 };
 
+/*
+ * Exit from all Iframes and go to the main document
+ *
+ * @param {Function} done
+ */
+
+exports.exitIframe = function(done){
+  debug('.exitIframe()');
+  this.evaluate_now(function() {
+      var element = __nightmare.rootDocument;
+      __nightmare.currentDocument = element;
+    }, done);
+};
+
 /**
  * Wait
  *
--- a/lib/actions.js
+++ b/lib/actions.js
@@ -280,6 +280,21 @@ exports.refresh = function(done) {
   }, done);
 };
 
+/*
+ * Enter to IFrame and run actions as if from inside
+ *
+ * @param {String} selector
+ * @param {Function} done
+ */
+
+exports.enterToIframe = function(selector, done){
+  debug('.enterToIframe() ' + selector);
+  this.evaluate_now(function(selector) {
+      var element = __nightmare.currentDocument.querySelector(selector).contentDocument;
+      __nightmare.currentDocument = element;
+    }, done, selector);
+};
+
 /**
  * Wait
  *
--- a/lib/actions.js
+++ b/lib/actions.js
@@ -287,8 +287,8 @@ exports.refresh = function(done) {
  * @param {Function} done
  */
 
-exports.enterToIframe = function(selector, done){
-  debug('.enterToIframe() ' + selector);
+exports.enterIframe = function(selector, done){
+  debug('.enterIframe() ' + selector);
   this.evaluate_now(function(selector) {
       var element = __nightmare.currentDocument.querySelector(selector).contentDocument;
       __nightmare.currentDocument = element;
--- a/lib/nightmare.js
+++ b/lib/nightmare.js
@@ -246,6 +246,9 @@ Nightmare.prototype.run = function(fn) {
  * you should not use this, unless you know what you're doing
  * it should be used for plugins and custom actions, not for
  * normal API usage
+ *
+ * It also enables running actions in inner frames by embedding action names
+ * with a custom document element.
  */
 
 Nightmare.prototype.evaluate_now = function(js_fn, done) {
@@ -259,7 +262,20 @@ Nightmare.prototype.evaluate_now = function(js_fn, done) {
   var args = Array.prototype.slice.call(arguments).slice(2);
   var argsList = JSON.stringify(args).slice(1,-1);
 
-  child.emit('javascript', template.execute({ src: String(js_fn), args: argsList }));
+  var sendJsFn = String(function(){
+    if (!__nightmare.currentDocument){
+      __nightmare.currentDocument = document;
+    }
+    return (function(){
+      var document = __nightmare.currentDocument;
+      return (js_fn)();
+    });
+  });
+
+  sendJsFn = sendJsFn.replace('js_fn', String(js_fn));
+
+  child.emit('javascript', template.execute({ src: sendJsFn, args: argsList }));
+
   return this;
 };
 
--- a/lib/nightmare.js
+++ b/lib/nightmare.js
@@ -263,10 +263,8 @@ Nightmare.prototype.evaluate_now = function(js_fn, done) {
   var argsList = JSON.stringify(args).slice(1,-1);
 
   var sendJsFn = String(function(){
-    return (function(){
-      var document = __nightmare.currentDocument;
-      return (js_fn).apply(this, arguments);
-    }.apply(this, arguments));
+    var document = __nightmare.currentDocument;
+    return (js_fn).apply(this, arguments);
   });
 
   sendJsFn = sendJsFn.replace('js_fn', String(js_fn));
--- a/lib/nightmare.js
+++ b/lib/nightmare.js
@@ -262,14 +262,14 @@ Nightmare.prototype.evaluate_now = function(js_fn, done) {
   var args = Array.prototype.slice.call(arguments).slice(2);
   var argsList = JSON.stringify(args).slice(1,-1);
 
-  var sendJsFn = String(function(){
+  var sendJsFn = String(function(argsList){
     if (!__nightmare.currentDocument){
       __nightmare.currentDocument = document;
     }
-    return (function(){
-      var document = __nightmare.currentDocument;
-      return (js_fn)();
-    });
+    return (function(argsList){
+      // var document = __nightmare.currentDocument;
+      return (js_fn)(argsList);
+    }(argsList));
   });
 
   sendJsFn = sendJsFn.replace('js_fn', String(js_fn));
--- a/lib/nightmare.js
+++ b/lib/nightmare.js
@@ -263,10 +263,6 @@ Nightmare.prototype.evaluate_now = function(js_fn, done) {
   var argsList = JSON.stringify(args).slice(1,-1);
 
   var sendJsFn = String(function(){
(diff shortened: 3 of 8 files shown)


tiangolo:
I see there are failing tests, I'm checking them (first setting up the environment to check them).

tiangolo:
* I fixed the errors
* refactored the creation of the variables to handle  `document` variables for iframe navigation (`__nightmare.currentDocument`)
* created fixtures for testing navigation with iframes
* created tests for navigation with iframes

erikdonohoo:
This would be super helpful for us since we use iframes in our widgets. :+1:

rclai:
When will this be reviewed/merged?

erikdonohoo:
I'm nervous that this project hasn't seen a commit in almost a month.  Its like it just died off completely.

tiangolo:
Me too @erikdonohoo. I guess Segmentio folks are too busy to develop the project further, but they could al least merge PRs...

tiangolo:
In the meantime, maybe you could try it @erikdonohoo and @rclai to see if you find any bug or misbehavior... 

You could clone my fork, checkout to the branch with the fix and copy the code to your project's `node_modules` directory to test it... I guess if someone from the Segmentio team sees that others have tested this PR and agree that it's working it will be easier for them to merge it....

rclai:
It works great @tiangolo, thanks for this.

rclai:
As I was working on [my issue](https://github.com/segmentio/nightmare/issues/536), I ran into this error:
```
Cannot read property 'contentDocument' of null
```
I'm thinking it's coming from [here](https://github.com/tiangolo/nightmare/blob/e1476d3f3f9dd41888f829818ca03ecf0bae933b/lib/actions.js#L293).

I got this error when I ran my script but using the Xvfb server instead of the `xvfb` NPM package.

```
# start xvfb server
export DISPLAY=':99.0'
Xvfb :99 -screen 0 1024x768x24 > /dev/null 2>&1 &
# Finally run my node script
node index.js
```

rclai:
Actually perhaps a better error should be thrown indicating that the iframe does not exist?

tiangolo:
@rclai I'm glad it's working!

---

For your issue / suggestion, I guess there are these approaches:

1. The user checks if the IFrame exists before entering in it, maybe using `.exists()`: https://github.com/segmentio/nightmare#existsselector

2. Nightmare checks if the IFrame exists before entering in it, and only proceeds if the IFrame exists.

3. Nightmare checks if the IFrame exists before entering in it and throws an error if it doesn't.

Nevertheless, I think none of the other methods check or throw any error, for example, if you run `nightmare.type('.weird .non .existent .selector', 'some text that never gets typed')` you don't receive any error either.

I think that options 1 or 2 would be the most similar to how the rest of the code works.

---

But now, I'm actually worried that no one from Segmentio has even entered the conversation, and it has been quite some time... :disappointed:  

...even though this seems a useful feature (at least for the three of us).

rclai:
Is it possible to turn it as a plugin? Or does it modify some internal stuff?

tiangolo:
@rclai: It modifies internal stuff (quite "core" internal stuff).

It changes the preloaded JS sent to the browser.

I actually first tried creating a plug-in, but it wouldn't work without changing the core, or I would have had to override every method (and every future method) in Nightmare...

BTW, debugging tip (related to the internal stuff): It saves the root document object in `__nightmare.rootDocument` and the current IFrame's document object in `__nightmare.currentDocument`.

So, you can check in your Developer Tools (the console in your Nightmare Chromium window) if your `__nightmare.currentDocument` is pointing to the document that you expect to be in.

mbernabeu:
Nice feature, works perfect

tiangolo:
Great! Thanks @mbernabeu for testing it!

yunicz:
merge this please, i really need the iframe support

rosshinkley:
@tiangolo I think this would be better served as a plugin for Nightmare.  Thoughts?

tiangolo:
@rosshinkley, I tried that first but it's impossible as it modifies internal stuff. 

Check the full answer to that same question 4 comments above: https://github.com/segmentio/nightmare/pull/496#issuecomment-197568154

rosshinkley:
@tiangolo I saw the comment, thanks for the pointer.

I'm not entirely convinced it's not possible to turn this into a plugin.  Unlike, say, popups that make a new `BrowserWindow` which won't have _any_ of the Nightmare sugar baked in, the IFrames you're dealing with should still be in the same context.  I can see with modifying preload this might not be possible.  I'd have to play around with this PR to be sure.

Even if it weren't possible, I would be more inclined to modify the current plugin functionality to allow for functions like this.

Thoughts?

Mr0grog:
Just my 2¢ as someone who is interested in this, but not currently hurting from a lack of it (i.e. please take this with a grain of salt):

Having this is super awesome (really!), BUT, reading the code, it seems like this will only really affect actions that use `evaluate` or `evaluate_now` under the hood. Even then, it only works for ones that confine themselves to `document`—so it won’t confine actions like `scrollTo` or `back` to the specified iframe because they use `window` (though that is an easy bug to fix here). `inject`, similarly, will have surprising results (in an iframe, the `document` jQuery is dealing with will be different from the `document` someone calling `evaluate` is dealing with).

It also might have surprising effects for someone using the `evaluate` action and touching the `window` object, using event constructors or other built-in functions, any globals, or, writing code that tests types (since code is running in the parent frame but meant to interact with content from the iframe, where constructor identities for things like `Array`, `Object`, etc. will be different).

And of course, does this work well with `type` and `insert`? (That’s a legit question; I’m really not sure.) It certainly *won’t* work with injecting CSS, `goto`, `size`, `userAgent`, `screenshot`, `html`, and `pdf`. Some of those are probably OK (I don’t think anyone would expect `size` to just resize the iframe), but others are more questionable. What should happen when I do:

```js
nightmare
  .goto(‘some_page_with_an_iframe')
  .enterIframe(‘#theframe')
    .inject('some_file.css')
    .inject(‘some_file.js')
    .html()
```

…and so on? If the current behavior with this page is OK, the fact that some actions will behave a little oddly should definitely be made clear in documentation.

Bottom-line, though, it seems to me that there are a lot of caveats to this that haven’t been talked through here.

tiangolo:
@rosshinkley, I tried at first all the ideas I had to make it a plug-in, but I had to throw all that code away and start from scratch because I found that I had to modify deeper functionality in Nightmare.

If you find a way to get that same behavior with a plug-in I would love to know how.

@Mr0grog that's a good point, not **every** action would work. I think the first approach would be to document which actions wouldn't work.

But at least for some cases, the actions that work seem to be quite useful for some of us.

I'm sure it works with `.type` (that's what I made it for).
I'm pretty sure it would also work for `.insert`, as the code is almost the same and also depends on `document`: https://github.com/segmentio/nightmare/blob/master/lib/actions.js#L175

Could you list all the *caveats* that you find that *haven’t been talked through here*? How would you like to assess better those caveats so that we all know what would have to be done?

rosshinkley:
I finally had some time to circle back to this.  The short version is you _can_ do this with a plugin, but  not without overriding the `evaluate_now` method on the prototype.

The `document` (really, `window.top.document`) variable isn't settable (which is I'm guessing why @tiangolo had trouble with the plugin route).  Because of that, you can't set the `document` variable from within a regular plugin method.  This has the consequence of the only place remaining to alias `document` - at least in the current implementation of Nightmare - is in `evaluate_now`.  You _can_ override the Nightmare prototype's `evaluate_now`, and I've thrown together [a (horrifying) sample](https://github.com/rosshinkley/nightmare-iframe-manager).  (Use at your own peril.)

Patching `evaluate_now` to wrap `document` _technically_ works, but suffers from the same issues that implementing frame support directly suffers from.  Off the top of my head:

* `.scrollTo()` only scrolls the window, not the interior selected frame.  Scrolling interior to an `iframe` would have to be written.
* `.screenshot()` will still take a shot of the bounding rectangle if provided, including the parent frame(s).  Not that this is unexpected, but it may catch new users off guard.
* `.html()` and `.pdf()` will still return output for the entire page rather than the selected child frame.  Again, expected behavior, but not what I'd expect as a new user.

There might be other issues with some of the more advanced functions (authentication, for example), but I haven't had a chance to set up testing yet.

rosshinkley:
I should have reread the _entire_ thread before posting: [@Mr0grog covers issues](https://github.com/segmentio/nightmare/pull/496#issuecomment-208445167) more comprehensively than I did.  (Price of being in a hurry, apologies.)  I still think there are almost certainly other corner cases, though.  As time permits, I'll set up some test cases to verify what @Mr0grog outlined.

Setting that aside, it frames need to be a first-order citizen in Nightmare (along with `BrowserWindow`s).  I think this should make it onto the list for vNext.

Mr0grog:
> Could you list all the caveats that you find that haven’t been talked through here?

- Globals variables (both built-ins and ones from scripts in the page in the frame). I think this is, far and away, the biggest thing. If people are expecting `evaluate` to execute code in the iframe, which is exactly the expectation this sets up, they would very reasonably expect to be able to access all the globals in that frame. They’d also expect to be able to be able to do things like compare constructors (e.g. `someFunctionInFrameThatReturnsArray().constructor !== Array`). There are an infinitely ridiculous number of use cases and ways people could trip over this.

- I think there’s a reasonable argument that it should probably work with `inject`. There are several issues related to users `inject`-ing jQuery (or a helper library of their choice) so they can use it when calling `evaluate`, so we know it’s something people are doing. On a cursory reading, it looks like it should *sort of* work with JS (caveat the global issues above), but won’t work with CSS.

- Obviously, a discussion of all the Nightmare methods that are/aren’t affected. @rosshinkley listed a few. If you’re going to address possible caveats, you should list them all out. Ross also noted “there might be other issues with some of the more advanced functions (authentication, for example), but I haven't had a chance to set up testing yet.” No offense to anyone here, but… testing and validating that shouldn’t be on Ross. There are *a lot* of issues and PRs here that need attention and Ross appears to be the only active maintainer.

- I think the issue of security is kind of a big deal here. The patch adds documentation that suggests setting `{'webPreferences': {'webSecurity': false}}` without really discussing that someone should at least think twice before doing that. There are a few issues that demonstrate people are using Nightmare to do tasks in very security-sensitive settings (e.g. #545); setting `websecurity: false` while doing something like that is a Really Bad Idea™. It may even be worth fixing #5115 upstream first to alleviate this.

Not a caveat, but probably also a good idea to have tests that cover all the other methods that are affected by this. e.g. `can type into elements in iframe`.

Other minor nit: maybe worth renaming to `enterFrame` since this is just as applicable to other kinds of frames? (`<frame>`, `<iframe>`, `<object>`)

And an open question, because there’s been no discussion of it here and I think it’s important for any new API being added to core: is this the best API for working with subframes?

- The original issue (#203) had an API using functions that could have eliminated the need for a second method call to “exit” the frame.
- If this is going into core, would it better to change the methods that take selectors to allow special selectors that reach into frames? e.g. `.click('.some > iframe / #a .button-in-the-frame')`

tiangolo:
Thank you @rosshinkley and @Mr0grog for paying attention to this.

I see that there are several issues that would be hard to deal with in trying to integrate this into Nightmare. And probably the way to go would be a new module as the one @rosshinkley built that patches `evaluate_now`.

Nevertheless, I'm being unable to use @rosshinkley's https://github.com/rosshinkley/nightmare-iframe-manager, its code never gets executed in the current environment. Maybe I'm doing something wrong, but I cloned the repo and tried to run the tests (after installing a couple undeclared packages) but they all seem to fail. Again, I'm probably doing something wrong, but maybe you can guide me on how we should use your package / module to get this feature.

rosshinkley:
@tiangolo I'd suggest opening issues at nightmare-iframe-manager.  (Looking at it offhand, I'd bet there's a problem with the Nightmare version.  I put the plugin together more as a proof of concept than anything.)  Time permitting, I'll see if I can fix the glaring problems.  If you could report the issue(s) you had there, that would help tremendously.

tiangolo:
OK, get it @rosshinkley. 

Are you planning on uploading it later to NPM? Do you want to keep it just as a proof of concept or do you want to keep developing it?
Or do you want me to fork it, try to fix my issues and then upload to NPM?

rosshinkley:
> Do you want to keep it just as a proof of concept or do you want to keep developing it?

I can keep developing it, but remember: it's going to be _very_ fragile.  Using it is probably not a good idea.

> Are you planning on uploading it later to NPM?

I was/am reluctant to do so because it modifies a core part (virtually _everything_ touches `evaluate_now()`) of how Nightmare works.  I don't know that I can make the disclaimer large enough.  I can publish it to NPM if it solves anything, but keeping it confined to Github keeps it in the realm of "you can, but probably shouldn't."

tiangolo:
For the other Nightmare users in this thread that wanted this functionality (with all the risky security concerns already mentioned), using @rosshinkley's plug-in is working now. 

His plug-in is based on this PR and even improves the functionality.

Check his documentation at: https://github.com/rosshinkley/nightmare-iframe-manager

(Make sure to update your code as the names and meaning of the methods changed a little bit).

As there isn't an NPM package, to install it in your projects you need to run:

```
npm install --save rosshinkley/nightmare-iframe-manager
```

NPM will understand that it needs to go for the GitHub repo (but you will need to have git installed).

---

@rosshinkley and @Mr0grog, with that plug-in I have the functionality I (and I think most of us in this thread) wanted (with all the risks / security up to us). For me it's OK if you just reject and close this PR now. Thanks!

Issue #505, opened 2016-02-25
How do I open DevTools?


moroshko:
Is there a built-in method to open DevTools for debugging purposes?

I tried to integrate [electron-debug](https://github.com/sindresorhus/electron-debug), but no luck (Electron was throwing errors).

rosshinkley:
I created #506 to show how you can open DevTools.

For more information, you may want to go have a look at the [`webContents` documentation](https://github.com/atom/electron/blob/master/docs/api/web-contents.md) for Electron, specifically the [DevTool specific parts](https://github.com/atom/electron/blob/master/docs/api/web-contents.md#webcontentsaddworkspacepath).

kokujin:
This is great, does that mean that we can use the 'detach' parameter too?

````
webContents.openDevTools([options])

options Object (optional)
detach Boolean - opens DevTools in a new window
`````

**Comment by rosshinkley** [comment]
Sure, I've updated #506 to reflect that.  You can now pass a hash in with `openDevTools` that get passed directly to `openDevTools()`.

**Comment by kokujin** [comment]
Thanks @rosshinkley, but the flag is just to open the devTools drawer but not detach it. Correct?

**Comment by rosshinkley** [comment]
@kokujin No, the latest update will accept either `true` (to show the devtools) or an options hash that gets passed directly to `webContents.openDevTools()`.  For example:

```js
var Nightmare = require('nightmare'),
  nightmare = new Nightmare({show:true, openDevTools: { detach: true } });
```
...would detach the devtools window.

moroshko:
@rosshinkley Thanks a lot for #506! Let's leave this issue open until #506 is merged. At least, for now, I can patch it in `node_modules` to debug things :)

kokujin:
Still cant get this to work, is this under 2.20 or another branch? Thanks

rosshinkley:
@kokujin #506 hasn't been accepted, and as such, is not a part of Nightmare.

If you absolutely need it, you could use the [branch on my fork](https://github.com/rosshinkley/nightmare/tree/505) instead.  For the nightmare dependency, you could have an entry in your `package.json` like:

```json
"nightmare":"git://github.com/rosshinkley/nightmare#505"
```

Alternatively, you could fork Nightmare yourself and pull the changes in from that branch on my fork.

kokujin:
Great! Thanks @rosshinkley

PR #506, opened 2016-02-25
added `openDevTools` to Nightmare options

rosshinkley:
Fixes (or at least shows) an example of how to address #505.


lib/runner.js
```diff
@@ -76,6 +76,13 @@ app.on('ready', function() {
      */
 
     win = new BrowserWindow(options);
+    if(options.show && options.openDevTools){
+      if(typeof(options.openDevTools) === 'object') {
```
justinmchase: parenthesis around `options.openDevTools` is not necessary

PR #512, opened 2016-02-26
Should capture error if evaluate fn is not a function.

### LINKED ISSUE #511: Should capture error if evaluate fn is not a function
**Test Case**

```js
// test/index.js
it('should capture invalid evaluate fn', function*() {
    var didFail = false;
    try {
      yield nightmare
        .goto(fixture('evaluation'))
        .evaluate('not_a_function');
    } catch (e) {
      didFail = true;
    }
    didFail.should.be.true;
  });
});
```

**Source Code**

```js
// lib/nightmare.js
exports.evaluate = function (fn/**, arg1, arg2..., done**/) {
  // ...
  if (typeof fn !== 'function') {
    return done(new Error('.evaluate() fn should be a function'));
  }
  // ...
};
```

Before the fix, the test case would hang util timeout.
Related: #513

fritx:
Fix #511

### DIFF
--- a/lib/actions.js
+++ b/lib/actions.js
@@ -392,6 +392,9 @@ exports.evaluate = function (fn/**, arg1, arg2..., done**/) {
   var args = sliced(arguments);
   var done = args[args.length-1];
   var newArgs = [fn, done].concat(args.slice(1,-1));
+  if (typeof fn !== 'function') {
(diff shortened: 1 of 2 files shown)


rosshinkley:
I like this change.  Marking for inclusion in the next release.

fritx:
Thanks!

PR #514, opened 2016-02-26
Should capture syntax error with injected js.

### LINKED ISSUE #513: Should capture syntax error with injected js
**Test Case**

``` js
// test/index.js
it('should inject javascript and return result', function*() {
  var result = yield nightmare
    .goto(fixture('manipulation'))
    .inject('js', 'test/files/eval.js');
  result.should.equal(6);
});

it('should inject javascript and capture syntax error', function*() {
  var didFail = false;
  try {
    yield nightmare
      .goto(fixture('manipulation'))
      .inject('js', 'test/files/error.js');
  } catch (e) {
    didFail = true;
  }
  didFail.should.be.true;
});
```

``` js
// test/files/error.js
/* syntax error */
{
```

``` js
// test/files/eval.js
/* expression to eval */
alert(123), 2*3
```

**Source Code**

``` js
// lib/nightmare.js
Nightmare.prototype._inject = function(js, done) {
  // ...
  // ~~child.emit('javascript', template.inject({ src: js }));~~
  child.emit('javascript', template.inject({ src: JSON.stringify(js) }));
  // ...
};
```

``` js
// lib/javascript.js
var inject = [
  "(function javascript () {",
  // ...
  "  try {",
  // ~~"    var response = (function () { {{!src}} })()",~~
  "    var response = eval({{!src}})",
  "    ipc.send('response', response);",
  "  } catch (e) {",
  "    ipc.send('error', e.message);",
  "  }",
  // ...
].join("\n");
```

**Note: this solution uses `eval`, but I guess it is safe cuz the `src` is `JSON.stringify()`d.**

Before the fix, the test case would hang util timeout.
Related: #511

fritx:
Fix #513


lib/javascript.js
```diff
@@ -43,7 +43,8 @@ var inject = [
   "    ipc.send('log', sliced(arguments));",
   "  }",
   "  try {",
-  "    var response = (function () { {{!src}} })()",
+  // "    var response = (function () { {{!src}} })()",
```
justinmchase: You should just delete removed code.

justinmchase:
What if they want to explicitly return?
```
if (some_thing)
  doSomStuf()
  return true
else
  return someOtherStuff()
```

fritx:
@justinmchase yeah, sorry for this.. this does make a change of the `inject` usage.

They have `result` returned with `func1();func2();result`.
If they want to have a function injected like yours, maybe they could use `evaluate` instead?

rosshinkley:
If you run Nightmare with `DEBUG` with your `error.js`, you will get a page error.  (It's true, though, the Nightmare process does hang as the `'javascript'` event is never emitted.)  If your `.inject()`ed code throws an exception, that gets bubbled up to the calling error handler.  If I'm reading this correctly, you want the syntax error to _also_ bubble up to the caller, correct?

If that's the case, I don't think you need to change how the injected script is executed, nor how the source is passed to the injection template.  You could modify the `'page'` handler in `runner.js` to something like:

```js
renderer.on('page', function(sender/*, arguments, ... */) {
      if(arguments[1] === 'error'){
        throw new Error(arguments[2]);
      }
      parent.emit.apply(parent, ['page'].concat(sliced(arguments, 1)));
    });
```

... which I think satisfies your requirements and tests.  However, this might have unintended side effects for someone working with a page that has errors internally.

Instead, you could add an _extremely_ permissive [`jslint`](https://github.com/reid/node-jslint/blob/master/doc/jslint.md) call to verify the `.inject()`ed file is valid javascript.

All of that said, I'm not sure I agree with the premise that `inject()`ed files should bubble up an exception, especially since they're under user control and the tools already exist to debug them.  Thoughts?

justinmchase:
Can we close this or accept it? I'm tired of seeing it in my PR list :)

PR #520, opened 2016-02-29
Expose media playback events

davidnaas:
Makes it possible to listen for `media-started-playing` and `media-paused` with the on method.

Useful in testing apps that deal with audio playback.


reinpk:
👍 thanks!

Issue #323, opened 2015-10-29
Can't inject minified jquery


chentsulin:
This works great:

```
nightmare.inject('js', `node_modules/jquery/dist/jquery.js`);
```

but minified file not:

```
nightmare.inject('js', `node_modules/jquery/dist/jquery.min.js`);
```

Any ideas?

TonyWang031:
Got the same issue too.

matthewmueller:
can you narrow down where the error might be occuring?

rosshinkley:
@chentsulin @TonyWang031 Depending on the version of jQuery being injected, they could end with sourcemap comments, which would cause the end of the calling function that encapsulates the injected script to be commented out, causing a runtime error.  I would be willing to bet this is what you were hitting, given the unminified version worked but the minified version wouldn't.

The template was fixed in #524, so I am going to go ahead and close this issue.  If you still have problems, feel free to reopen/submit a new one.

PR #524, opened 2016-03-01
remove double-injection from `.inject()`

### LINKED ISSUE #523: console.log messages from injected script doubles
I have two questions about script injections. 

The first one looks like a bug. 
Just use this pretty simple nightmare-script...
```javascript

var nightmare = new Nightmare()
    .on('console', function (type, message, errorStack) {
        if (type === 'log') {
            console.log(message);
        }
    })
    .goto('http://google.com')
    .inject('js', '_testScripts.js')
    .wait(1000)
    .end()
    .then();
```
...with this script for inject (*_testScripts.js)
```javascript
setTimeout(function() {
	console.log('Hello from script');
}, 0);
```

Result is: 
```
Hello from script
Hello from script
```

* * *
Second question exactly about the example: So, why if I use ```console.log('Hello from script');``` it doesn't work properly in the contrast of ```setTimeout(function() {console.log('Hello from script');}, 0);``` ?

rosshinkley:
Fixes #523 (if #476 is also included).

I'm not _real_ clear on why the injection was done twice to begin with, and perhaps @reinpk could shed some light on that?

On top of fixing that, I went through every version of jQuery core to determine which work with `.inject()` and which don't.  This may help address #323.  The following versions of jQuery _are not_ injectable with Nightmare (most throw syntax errors):

- 1.0.1
- 1.0.2
- 1.0.3
- 1.0.4
- 1.1.1
- 1.1.2
- 1.1.3
- 1.11.0 beta 1 min
- 1.9.0 min
- 2.1.0 beta 1 min

Other versions seem to work with basic queries.


Mr0grog:
> The following versions of jQuery are not injectable with Nightmare (most throw syntax errors)

All these version need to be looked at, but I suspect they might share similar issues to 1.9.0.min: it ends in a line comment for a sourcemap like so:

```js
blahblahblahminifiedunreadablecode;
//@ sourceMappingURL=jquery.min.map
```

And it just so happens that our injection template has code after the injected source on the same line (https://github.com/segmentio/nightmare/blob/master/lib/javascript.js#L46):

```js
var response = (function () { {{!src}} })()
```

So the fix is as simple as inserting a newline after the injected file’s source code:

```js
var response = (function () { {{!src}}\n })()
```

Kinda ridiculous, but works great and is generally more safe!

Mr0grog:
…after a very quick deeper look:

- `1.11.0 beta 1 min` and `2.1.0 beta 1 min` have the same issue as `1.9.0.min` and are fixed by the newline as noted above.
- `1.1.3` and older expect to be loaded and run in global scope. The just define `var $ = ...` instead of `window.$ = ...`. Since the injection template wraps all the code in a function, that variable never escapes the function’s scope and gets attached to the global object. That’s a more fundamental incompatibility; “fixing” older jQuery (and other scripts that behave like it) would necessarily break other features.

rosshinkley:
> So the fix is as simple as inserting a newline after the injected file’s source code:

@Mr0grog Ah, thank you.  I haven't had time to double back and take a look at why the versions that are excepting were causing syntax errors.

> So the fix is as simple as inserting a newline after the injected file’s source code:

Simple, yes - this makes me wonder how common a practice it is to _end_ a minified file with a sourcemap or comment.  It's out of scope for this question, but I'm still curious.

At any rate, I'll add the newline into the template .  Changes shortly.

> 1.11.0 beta 1 min and 2.1.0 beta 1 min have the same issue as 1.9.0.min and are fixed by the newline as noted above.

Good to know.  I think 1.9.x is probably the biggest offender.

> 1.1.3 and older expect to be loaded and run in global scope. They just define var $ = ... instead of window.$ = .... Since the injection template wraps all the code in a function, that variable never escapes the function’s scope and gets attached to the global object. That’s a more fundamental incompatibility; “fixing” older jQuery (and other scripts that behave like it) would necessarily break other features.

I'm unsurprised.  I vote leaving <1.1.x broken until compelling case to fix it is made.

That said, I wonder what use case exists where one would be purposefully injecting jQuery <1.1.x with Nightmare.  I could understand pages that already have early versions of jQuery included, but I can't think of a good reason to inject an old version other than for an experiment to determine what Nightmare works with.

Mr0grog:
> this makes me wonder how common a practice it is to end a minified file with a sourcemap or comment.

Totally speculative, but I’ll bet it’s only becoming more common with modern tooling. It’s good unixy practice to always end files with a line break (which would have avoided this issue), but that’s obviously not something we can rely on.

> I vote leaving <1.1.x broken until compelling case to fix it is made.

At the very least, a change to support <1.1.x would be a big enough change I think it’s out of scope here, so from that perspective, I agree :)

> I wonder what use case exists where one would be purposefully injecting jQuery <1.1.x with Nightmare.

I doubt there is one (or at least that there is one that’s particularly compelling). However, plenty of other libraries could behave the same way (which is, of course, hypothetical until we hit one that’s modern and does, which again is unlikely since I think most libraries today will use some UMD-esque syntax, which would generally not run afoul of this issue).

That said, I can definitely see an expectation on the part of the user that:

`inject('js', 'path/to/file.js')`

would behave just like

```html
<script src="path/to/file/if/only/it/was/being/served/over/http.js"></script> or
<script>[contents of file]</script>
```

in which case a script that behaves like old jQuery would work. Maybe worth a “known issues” document over at https://github.com/rosshinkley/nightmare-examples

rosshinkley:
@Mr0grog A bit late, but I've added [a first draft](https://github.com/rosshinkley/nightmare-examples/blob/master/docs/known-issues/globally-defined-variables.md) of the globally defined variables example.

PR #529, opened 2016-03-04
Allow for audio muted state to be set on the electron process.

Oceanswave:
Audio may be required to be played in certain circumstances, this allows
the default audio mute state to be overridden.


Oceanswave:
Ug. getting random CI test failures from unrelated code.

PR #475, opened 2016-02-12
Add authentication

brrd:
Electron now supports basic authentication : https://github.com/atom/electron/pull/3250
This PR adds an `.authentication()` method to Nightmare.

PR #538, opened 2016-03-19
Support HTTP Basic Authentication

neekolas:
Adds support for 'login' event.
See: https://github.com/atom/electron/pull/3250

Adds a new option defaultAuthentication

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -144,6 +144,15 @@ var nightmare = Nightmare({
 });
 ```
 
+##### defaultAuth
+A username and password pair for HTTP basic auth on all requests
+
+```js
+var nightmare = Nightmare({
+  defaultAuth: 'username:password'
+});
+```
+
 #### .useragent(useragent)
 Set the `useragent` used by electron.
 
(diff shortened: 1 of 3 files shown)


Readme.md
```diff
@@ -144,6 +144,15 @@ var nightmare = Nightmare({
 });
 ```
 
+##### defaultAuth
+A username and password pair for HTTP basic auth on all requests
+
+```js
+var nightmare = Nightmare({
+  defaultAuth: 'username:password'
```
justinmchase: I feel like this might be better as a more nested object like:
```
auth: {
  basic: {
    user: '...',
    pass: '...'
  }
}
```

rosshinkley:
I think this is a good inclusion, provided it's unit tested.  I think using something like [`basic-auth`](https://www.npmjs.com/package/basic-auth) with the test Express server would work.

brrd:
Looks like a duplicate #475

justinmchase:
#475 looks like a simpler version of this with tests already.

rosshinkley:
@brrd I'm sorry, I must have missed #475.

@justinmchase Agreed, closing as duplicate.

PR #548, opened 2016-03-30
blur text inputs when finished with type() or insert()

yoz:
I was inspired to do this because some event listeners e.g. 'change' are triggered on blur, so they don't happen until we actually click/focus another element.

P.S. insert and type are very similar - there's some potential for refactoring but I don't know what style we would prefer.

### DIFF
--- a/lib/actions.js
+++ b/lib/actions.js
@@ -155,16 +155,17 @@ exports.type = function() {
   debug('.type() %s into %s', text, selector);
   var child = this.child;
 
-  if((text || '') == '') {
-    this.evaluate_now(function(selector) {
-      document.querySelector(selector).value = '';
-    }, done, selector);
-  } else {
-    focusSelector.bind(this)(function() {
-      child.once('type', blurSelector.bind(this, done, selector));
+  focusSelector.bind(this)(function() {
+    var blurDone = blurSelector.bind(this, done, selector);
+    if ((text || '') == '') {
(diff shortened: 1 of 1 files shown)


yoz:
Although some of the tests are timing out on CircleCI, it looks like that's an issue with flaky tests rather than with my change (tests pass for me locally).

rosshinkley:
I'm behind this change.  I can't see any downside to adding blur events to both.

I do have a question, and I realize this is how it currently works - when empty string or a falsey value is passed, the selector's value is set to empty string, but no focus or blur actions are taken.  Won't that cause problems if the intent is to clear the selector and the change event is triggered on blur?  Shouldn't clear events also cause blur?

Re the tests, I kicked off another build, and it passed.  The flaky tests might be a result of event conditions like those found in #553.  I'm attempting to get to the bottom of it.

rosshinkley:
@yoz Any updates?

By the by, the tests were fixed with a makefile/shell script stopgap in #565.

yoz:
@rosshinkley Sorry, got caught up in other work - I think it makes sense to add focus/blur events when clearing inputs, too. Will hopefully get to that later today.

rosshinkley:
@yoz I understand.  Thanks for putting it together!

yoz:
Updated to trigger focus/blur events when we clear inputs.

Issue #527, opened 2016-03-03
Difference between `nightmare.then` and `promise.then`


Mr0grog:
In attempting to explain the behavior of `run` and `then` for #516, I realized the details of how Nightmare’s `then` method works means you can’t just say “works like `promise.then`.” The difference is minor and the likelihood of it tripping someone up is probably low, but it could still happen, especially given the promise-y expectations the method’s name sets up. Since `then` is now the preferred API, it should really be documented and, in doing so, the details of `then` should be made clear.

In the spirit of @rosshinkley’s comments on #491:

> I am certain it's going to come up in questions again and I wanted to have a reference to point at.

It might be helpful to understand *why* Nightmare’s `then` method does not behave in accordance with a promise’s `then` (or any Promises/A+ compliant thenable) in order to better document it.

Specifically, *if an error occurs in the success callback that is passed as the first argument to `then` (or if that callback returns a promise that rejects), the error callback passed as the second argument to `then` will handle it instead of it bubbling up the promise chain.*

Is this difference intentional? If so, what was the intent behind that decision? What was it meant to enable or prevent, if anything? If not, is this a bug to be fixed?

rosshinkley:
I suspect the difference is a side effect of ensuring `vo`, `co`, and possibly `mocha-generators` Just Work&trade;.  Consider the following (admittedly contrived) example:

```js
var Nightmare = require('nightmare'),
    vo = require('vo');

function * run() {
    var nightmare = new Nightmare({
        show: false,
    });

    yield nightmare
        .goto('http://example.com')
        .wait('body')
        .evaluate(function(){
            throw new Error('uh oh');
        });

    yield nightmare.end();
}

vo(run)(function(err) {
    if (err) {
        console.dir(err);
    }
    console.log('done');
});
```

Out of the box, `vo` runs the promise and passes internal methods to `fulfill` and `reject` as prescribed by [the native promise documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise).  If you remove the `.catch()` call from Nightmare under `.then()`, the above example hangs.  Simillarly, with `co`, the rejection callback is never called (for hopefully obvious reasons).

All of that said, I agree with you that it is a _little_ misleading that you couldn't modify my original example to something like the following:

```js
var Nightmare = require('nightmare'),
var nightmare = Nightmare();
nightmare
  .goto('http://example.com')
  .wait('body')
  .evaluate(function() {
[...]

Issue #493, opened 2016-02-19
Simultaneous calls to the same action can cause early returns/continues with the wrong results


Mr0grog:
Originally found as a byproduct of #465, also noted over in the discussion on #491.

If two queues on the same Nightmare instance are operating simultaneously and both call the same action (e.g. `evaluate()`), it is possible for one call to pick up the results from the first call and finish early, continuing its cue with the wrong data.

Here's an example from #465:

``` js
var Nightmare = require('nightmare');
var page = Nightmare({show:true}).goto('http://yahoo.com');

page
  .wait(function() {
    return false;
  })
  .run(function(error, result) {
    if (error) { console.log("Error waiting for nothing:", error); }
    else { console.log("Inifnite wait finished, but shouldn't have! (uhoh)"); }
  });

page
  .wait(function() {
    return document.querySelector(".darla").firstChild.nodeName === "IFRAME";
  })
  .run(function(error, result) {
    if (error) { console.log("Error waiting for ad to load:", error); }
    else { console.log("First ad is loaded"); }
  });

// Prints:
// First ad is loaded
// Inifnite wait finished, but shouldn't have! (uhoh)
```

Or with typing:

``` js
var Nightmare = require('nightmare');
var page = Nightmare({show:true}).goto('http://yahoo.com').run(function() {
  page
    .type('#uh-search-box', 'github nightmare')
    .evaluate(function() {
      return document.querySelector('#uh-search-box').value;
    })
    .run(function(error, result) {
      if (error) { console.log("Error typing:", error); }
      else { console.log("Typed 'github nightmare'? " + result); }
    });

  page
    .type('#uh-search-box', 'ok')
    .evaluate(function() {
      return document.querySelector('#uh-search-box').value;
    })
    .run(function(error, result) {
      if (error) { console.log("Error typing:", error); }
      else { console.log("Typed 'ok'? " + result); }
    });
});

// Prints:
// Typed 'github nightmare'? goikt
// Typed 'ok'? goikt
```

Obviously there are more complicated questions with that one, but the salient bit here is that the queue typing "github nightmare" returned early. You can watch the window and see that it is still typing after the results are printed to the console.

You can do this with `goto()`, too, but that one's also somewhat complicated by the fact that `goto()` actually doesn't know how to fail (which is a different issue altogether). But it would fail to fail (ha!) if navigation were pre-empted by a second, simultaneous `goto()` call.

[...]

PR #553, opened 2016-04-04
Goto should fail if the page doesn’t load

Mr0grog:
~~**This is just a work-in-progress at the moment, but filing the PR as there are open questions…**~~

This adds tests (which fail), but no implementation—I need to work out the best solution for https://github.com/atom/electron/issues/5013 and https://github.com/atom/electron/issues/4396 (this second one I can’t actually reproduce, so it may have been fixed). For the former, the simplest solution is probably tracking the URL of the failed page and comparing it to the URL of the main frame. Need to make sure that works right for redirects, though. A more exacting solution might be to track resource requests on the `session` object, though doing so requires a bit more rigamarole.

There is also an open question of how this should behave w/r/t error responses (e.g. 404, 500, etc). For now, I’m planning to take the approach that any valid response (including things like 404s and 500s) are successes (after all, you might be trying to test a 404 page with Nightmare), but I can also imagine someone wanting the option to treat any non-2xx response as a failure. Maybe that’s something to wait for as a feature request, though? Curious what others think.

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -168,6 +168,23 @@ Complete any queue operations, disconnect and close the electron process.
 #### .goto(url)
 Load the page at `url`.
 
+When a page load is successful, `goto` returns an object with metadata about the page load, including:
+
+- `url`: The URL that was loaded
+- `code`: The HTTP status code (e.g. 200, 404, 500)
+- `method`: The HTTP method used (e.g. "GET", "POST")
+- `referrer`: The page that the window was displaying prior to this load or an empty string if this is the first page load.
+- `headers`: An object representing the response headers for the request as in `{header1-name: header1-value, header2-name: header2-value}`
+
+If the page load fails, the error will be an object wit the following properties:
+
+- `message`: A string describing the type of error
+- `code`: The underlying error code describing what went wrong. Note this is NOT the HTTP status code. For possible values, see https://code.google.com/p/chromium/codesearch#chromium/src/net/base/net_error_list.h
+- `details`: A string with additional details about the error. This may be null or an empty string.
+- `url`: The URL that failed to load
+
+Note that any valid response from a server is considered “successful.” That means things like 404 “not found” errors are successful results for `goto`. Only things that would cause no page to appear in the browser window, such as no server responding at the given address, the server hanging up in the middle of a response, or invalid URLs, are errors.
+
 #### .back()
 Go back to the previous page.
 
--- a/lib/nightmare.js
+++ b/lib/nightmare.js
@@ -144,6 +144,7 @@ function Nightmare(options) {
   // proporate events through to debugging
   this.child.on('did-finish-load', function () { log('did-finish-load', JSON.stringify(Array.prototype.slice.call(arguments))); });
   this.child.on('did-fail-load', function () { log('did-fail-load', JSON.stringify(Array.prototype.slice.call(arguments))); });
+  this.child.on('did-fail-provisional-load', function () { log('did-fail-provisional-load', JSON.stringify(Array.prototype.slice.call(arguments))); });
   this.child.on('did-frame-finish-load', function () { log('did-frame-finish-load', JSON.stringify(Array.prototype.slice.call(arguments))); });
   this.child.on('did-start-loading', function () { log('did-start-loading', JSON.stringify(Array.prototype.slice.call(arguments))); });
   this.child.on('did-stop-loading', function () { log('did-stop-loading', JSON.stringify(Array.prototype.slice.call(arguments))); });
--- a/lib/runner.js
+++ b/lib/runner.js
@@ -3,7 +3,8 @@
  */
 
 var parent = require('./ipc')(process);
-var BrowserWindow = require('electron').BrowserWindow;
+var electron = require('electron');
+var BrowserWindow = electron.BrowserWindow;
 var defaults = require('deep-defaults');
 var assign = require('object-assign');
 var join = require('path').join;
@@ -11,7 +12,10 @@ var sliced = require('sliced');
 var renderer = require('electron').ipcMain;
 var app = require('electron').app;
 var fs = require('fs');
+var urlFormat = require('url');
 var FrameManager = require('./frame-manager');
+
+const KNOWN_PROTOCOLS = ['http', 'https', 'file', 'about'];
 
 /**
  * Handle uncaught exceptions in the main electron process
@@ -225,11 +229,33 @@ app.on('ready', function() {
         parent.emit('goto', error, data);
       }
 
-      win.webContents.on('did-fail-load', handleFailure);
-      win.webContents.on('did-get-response-details', handleDetails);
-      win.webContents.on('did-finish-load', handleFinish);
-      win.webContents.loadURL(url, {
-        extraHeaders: extraHeaders
+      // In most environments, loadURL handles this logic for us, but in some
+      // it just hangs for unhandled protocols. Mitigate by checking ourselves.
+      function canLoadProtocol(url, callback) {
+        var protocol = (urlFormat.parse(url).protocol || '').replace(/:$/, '');
+        if (!protocol || KNOWN_PROTOCOLS.includes(protocol)) {
+          return callback(true);
+        }
+        electron.protocol.isProtocolHandled(protocol, callback);
+      }
+
+      canLoadProtocol(url, function(canLoad) {
+        if (canLoad) {
+          win.webContents.on('did-fail-load', handleFailure);
+          win.webContents.on('did-get-response-details', handleDetails);
+          win.webContents.on('did-finish-load', handleFinish);
+          win.webContents.loadURL(url, {
+            extraHeaders: extraHeaders
+          });
+          return;
+        }
+
+        done({
+          message: 'navigation error',
+          code: -1000,
+          details: 'unhandled protocol',
+          url: url
+        });
       });
     }
   });
--- a/lib/runner.js
+++ b/lib/runner.js
@@ -479,22 +479,3 @@ function forward(event) {
     parent.emit.apply(parent, [event].concat(sliced(arguments)));
   };
 }
-
-/**
- * URL pattern matching
- *
- * This is something of a poor man's imitation of Electron's URL matcher, which
- * is much more complicated:
- * https://github.com/electron/electron/blob/master/chromium_src/extensions/common/url_pattern.cc
- * It should work with most patterns designed for Electron's matcher, though.
- */
-function patternToRegExp(pattern) {
-  const source = pattern.split('*').map(function(part, index) {
-    return (index %2) ? '.*' : escapeRegExp(part);
-  }).join('');
-  return new RegExp('^' + source);
-}
-
-function escapeRegExp(text) {
-  return text.replace(/[.?*+^$[\]\\(){}|-]/g, '\\$&');
-}
--- a/lib/runner.js
+++ b/lib/runner.js
@@ -60,7 +60,7 @@ if (!processArgs.dock && app.dock) {
  */
 
 app.on('ready', function() {
-  var win, frameManager, options, currentLoadingUrl, startedLoading;
+  var win, frameManager, options;
 
   /**
    * create a browser window
@@ -130,43 +130,6 @@ app.on('ready', function() {
     win.webContents.on('plugin-crashed', forward('plugin-crashed'));
     win.webContents.on('destroyed', forward('destroyed'));
 
-    // `goto` needs to be able to listen for new requests starting in order to
-    // determine the actual URL that is being loaded (Electron will 'clean up'
-    // the URL under the hood, so the URL string passed to `goto` may not match
-    // the actual requests being made).
-    // Because webRequest allows only one handler at a time, we have to protect
-    // against a plugin disrupting our handler, so replace `onBeforeRequest`
-    // with a function that mimics the built-in behavior.
-    var webRequest = win.webContents.session.webRequest;
-    // If a request handler is registered, `beforeRequestHandler` will run it
-    // and return `true`. If none is present, return `false`.
-    var beforeRequestHandler = () => false;
-    webRequest.onBeforeRequest(function(details, callback) {
-      if (details.resourceType === 'mainFrame') {
-        startedLoading = true;
-        currentLoadingUrl = details.url;
-      }
-      beforeRequestHandler(details, callback) || callback({cancel: false});
-    });
-    webRequest.onBeforeRequest = function(filters, handler) {
-      if (arguments.length < 2) {
-        handler = filters;
-        filters = [];
-      }
-
-      if (!handler) { return () => false; }
-
-      var patterns = filters.map(patternToRegExp);
-      beforeRequestHandler = function(details, callback) {
-        var url = details.url;
-        if (!patterns.length || patterns.some(pattern => pattern.test(url))) {
-          handler(details, callback);
-          return true;
-        }
-        return false;
-      };
-    }
-
     parent.emit('browser-initialize');
   });
 
@@ -187,14 +150,10 @@ app.on('ready', function() {
     if (win.webContents.getURL() == url) {
       parent.emit('goto');
     } else {
-      currentLoadingUrl = url;
-      startedLoading = false;
       var responseData = {};
 
-      function handleFailure(event, code, detail, failedUrl) {
-        // No indicator for whether this is from the main frame, so use the URL
-        // as a proxy. See: https://github.com/electron/electron/issues/5013
-        if (!startedLoading || failedUrl === currentLoadingUrl) {
+      function handleFailure(event, code, detail, failedUrl, isMainFrame) {
+        if (isMainFrame) {
           done({
             message: 'navigation error',
             code: code,
@@ -205,8 +164,8 @@ app.on('ready', function() {
       }
 
       function handleDetails(
-        event, status, newUrl, oldUrl, statusCode, method, referrer, headers) {
-        if (newUrl === currentLoadingUrl) {
+        event, status, newUrl, oldUrl, statusCode, method, referrer, headers, resourceType) {
+        if (resourceType === 'mainFrame') {
           responseData = {
             url: newUrl,
             code: statusCode,
--- a/lib/runner.js
+++ b/lib/runner.js
@@ -56,7 +56,7 @@ if (!processArgs.dock && app.dock) {
  */
 
 app.on('ready', function() {
-  var win, frameManager, options;
+  var win, frameManager, options, currentLoadingUrl, startedLoading;
 
   /**
    * create a browser window
@@ -112,6 +112,7 @@ app.on('ready', function() {
 
     win.webContents.on('did-finish-load', forward('did-finish-load'));
     win.webContents.on('did-fail-load', forward('did-fail-load'));
+    win.webContents.on('did-fail-provisional-load', forward('did-fail-provisional-load'));
     win.webContents.on('did-frame-finish-load', forward('did-frame-finish-load'));
     win.webContents.on('did-start-loading', forward('did-start-loading'));
     win.webContents.on('did-stop-loading', forward('did-stop-loading'));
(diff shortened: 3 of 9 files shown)


Mr0grog:
Also: As you can see in the tests, I’ve made the assumption here that child frames and sub-resources (images, scripts, stylesheets) failing do not cause `goto` to fail (i.e. it only fails if the HTML itself fails to load). I’m pretty sure that’s the right thing, but happy to take feedback on that.

rosshinkley:
> This is just a work-in-progress at the moment, but filing the PR as there are open questions…

I'd like to take a crack at responding considering I +1'd the original issue.  I can't guarantee it'll be _good_ responses; most of this is pondering out loud, coffee in hand.  :)

> For the former, the simplest solution is probably tracking the URL of the failed page and comparing it to the URL of the main frame.

This (at least superficially) seems reasonable, at least until electron/electron#5013 is fixed.

> Need to make sure that works right for redirects, though.

I'd imagine you'd have to look for redirects and set the URL you'd be checking against appropriately.  Child resource redirects should be handled as `did-get-redirect-request` already has `isMainFrame` in the event data.  I don't _think_ you'd end up with a race condition as main frame page loads should not fail _after_ a redirect.  

I have a couple of questions, though.  What if the end user wants to do something custom with redirects - would this effect the ability to handle redirects with a custom action?  Would a (optional) destination assertion or command need to be included?  I'm thinking of cases where you want to explicitly want to check the URL after `.goto()` is executed - you could certainly do it with `.evaluate()`, but it might be nice to include a shorthand.

> A more exacting solution might be to track resource requests on the session object, though doing so requires a bit more rigamarole.

The best thing this would buy you is the filtering mechanism.  However, memory serving, filters aren't terrifically strict.  I don't think you can specify exact URL matches for filters.  Before I take a firm stand on this, though, I think it would be prudent to verify that assertion.  

The bigger problem with using `session.webRequest.onErrorOccurred()` - which admittedly, I'm making the huge assumption that's what you'd use - is I think it gets emitted when there is _any_ problem, not just page load failures.  I'm guessing this is the "rigamarole" you're referring to.  With a cursory look at the documentation, I'm not sure the event data would contain everything you'd need to make the distinction.  It's probably worth carving out time for the experiment, though.  I don't know enough about it to be certain.

> There is also an open question of how this should behave w/r/t error responses (e.g. 404, 500, etc).

I would posit that successfully loaded pages with HTTP error statuses are still successful loads for precisely the reason you cite: it's a legitimate test case.  If you needed to treat non-1xx/2xx/3xx as failures, you could hook into [`did-get-response-details`](https://github.com/electron/electron/blob/master/docs/api/web-contents.md#event-did-get-response-details).  Granted, this doesn't give you the ability to handle those responses directly in the Nightmare chain...

> Maybe that’s something to wait for as a feature request, though?

... and continuing my line of thought, I'm curious what you would expect usage to look like.  Are you envisioning an optional parameter for `.goto()`?  Something like `.goto(url[, status])`?  Or does this get back to the shorthand I mentioned above with some sort of destination filter/rule?  I don't know this needs to be solved right now, but it might be worth thinking about.

> Also: As you can see in the tests, I’ve made the assumption here that child frames and sub-resources (images, scripts, stylesheets) failing do not cause goto to fail (i.e. it only fails if the HTML itself fails to load). I’m pretty sure that’s the right thing, but happy to take feedback on that.

I'd say this is a valid assumption and agree.  I don't think internal page errors should explicitly cause Nightmare failures.  I would want Nightmare to behave as close to Real Browser&trade; would behave - as fault tolerant as possible, loading and running as much of a page as it can.  You could always add fail-fast behavior yourself, hooking into the failure events and throwing exceptions.


Thoughts?

Mr0grog:
Spent more time than I expected working through the technical issues (and adding a whole bunch more tests as a result) last night. Going to try and clean it up this evening when I have more time and push it up here.

> What if the end user wants to do something custom with redirects - would this effect the ability to handle redirects with a custom action?

I don’t believe it would, since you can’t manipulate the request from that event (you could do something that would abort and start a new one, but we’d be able to see that). That said, I wound up not using `did-get-redirect-request` for a *different* reason, so this probably doesn’t matter much anyhow.

(Now using `webRequest.onBeforeRequest`, which requires some working around for safety—only one handler can be registered at a time, which means we could clobber a plugin or a plugin could clobber us. But this lets us figure out the *actual* URL being requested, which may not match the URL a user passed in — `http://somewhere.com` gets requested as `http://somewhere.com/`, for example. Header redirects will cause `onBeforeRequest` to fire again, so the `did-get-redirect-request` becomes redundant.)

> The bigger problem with using session.webRequest.onErrorOccurred()… is I think it gets emitted when there is any problem, not just page load failures.

Well, any problem with a resource that has started loading, which, it turns out, doesn’t cover everything because some load failures pre-empt actual requests (e.g. totally invalid URLs). Aside from that, though, you can sort things out pretty easily w/ the `resourceType` parameter, which is `mainFrame` for the page itself being loaded in the main frame.

> Are you envisioning an optional parameter for .goto()? Something like .goto(url[, status])? Or does this get back to the shorthand I mentioned above with some sort of destination filter/rule?

I was originally thinking “Something like .goto(url[, status])” BUT had a more interesting thought last night, which was this: `goto` should return information about the response when it’s successful (probably using `did-get-response-details` as you mentioned). Filtering acceptable page loads could be another action (perhaps a plugin!) that comes next, e.g:

```js
nightmare
  .goto(‘somewhere')
  .withStatus(200)
  .evaluate(...)
  .etc;
```

Mr0grog:
OK! This works and *seems* pretty reliable now. As an interesting side note, this is the first time the [not-quite-normal-promise behavior](https://github.com/segmentio/nightmare/issues/527) has caught me in practice! (https://github.com/segmentio/nightmare/pull/553/commits/9acfa0dc4f2616e48748512373e3306065eaafa5#diff-910eb6f57886ca16c136101fb1699231R169)

---

## High level overview

`goto` now produces an error in the page fails to actually load. Completed requests with error *statuses* (e.g. 404, 500) are successes, not errors in this situation.

When there is an error, `goto` returns an object with the following:

- message: The string 'navigation error'
- code: Chromium's error code describing what went wrong. Note this is *NOT* the HTTP status code. For more details, see https://code.google.com/p/chromium/codesearch#chromium/src/net/base/net_error_list.h
- details: A string with additional details about the error. This may be null or an empty string.
- url: The URL that failed to load

When `goto` succeeds, it also now returns additional information about the request, so the user or other actions or plugins can do something meaningful with it. Success data is an object with the following:

- url: The URL that was loaded
- code: The HTTP status code (e.g. 200, 404, 500)
- method: The HTTP method used (e.g. "GET", "POST")
- referrer: The page that the window was displaying prior to this load or an empty string if this is the first page load.
- headers: An object where the property names are the HTTP header names and the values are arrays of the header values (if the same header was sent multiple times, the array will have multiple values).

I’m not totally sure if using `code` for both Chromium’s error code on a failure and the HTTP status code on a success is right. It seemed straightforward and expedient, but my one concern is that someone might expect `code` in an error to be an HTTP status code, which it most definitely is not. However, you can easily differentiate it by the fact that Chromium error codes are all negative values.

## Implementation Bits

There are a lot of tests for all the crazy edge cases I encountered trying to solve this well. Hopefully they are clear.

Additionally, there’s some serious craziness around `webRequest.onBeforeRequest()`. Between the fact that calling that method with a new listener (or no listener) removes any existing listener and the fact that plugins now have access to the Electron process, a plugin could easily clobber the mechanisms `goto` relies on (or vice versa). So I had to do some wacky gymnastics to protect against that.

An `onBeforeRequest` listener is registered immediately when the browser window is created. Then the `onBeforeRequest` method is removed so nobody can clobber the listener that was just registered. It gets replaced by a method with the same signature that tracks any new listener and calls it as appropriate, rather than actually registering it with Electron. The biggest downside here is that Electron’s `filters` parameter is rather poorly approximated by my code. [The actual implementation is not small or simple](https://github.com/electron/electron/blob/master/chromium_src/extensions/common/url_pattern.cc) and this seemed good enough. Would love to hear other better ideas here.

I also thought about replacing the whole of `session.webRequest` with an event emitter, but that seemed like a change that was a) a little large/out-of-scope and b) is questionable in general because it would totally change a big piece of Electron API—it might be better to lobby for the change upstream (but no idea how well a request for that would be received).

Anyway, why is all that hijinx really necessary?? Well, the short answer is that the only way to tell whether a `did-fail-load` event came from the main frame (as a opposed to a sub-frame, which usually means an `<iframe>`) is to compare URLs. And determining the actual URL we’re trying to load is not as straightforward as someone as naïve as me might assume. (Note: I wrote a PR upstream in Electron to give us frame information on the failure event, but no idea if/when it will land: https://github.com/electron/electron/pull/5029)

So, URLs! Between the time we call `loadURL()` and when Electron attempts to make an actual request, it cleans up and reformats the URL—which means we can’t just match the failed URL agains the URL we asked to load. Some examples of how the URL gets modified:

- `   http://fake.domain/   `  ->  `http://fake.domain/`
- `http://///fake.domain/`  ->  `http://fake.domain/`
- `http:fake.domain/`  ->  `http://fake.domain/`
- `http://fake.domain`  ->  `http://fake.domain/`
- `http:// fake.domain`  ->  `http://%20fake.domain/`

Those are just ones I noticed in the course of doing this work. It would probably take a fair bit of Spelunking through Chromium to figure out everything it might do and, even then, we’d be vulnerable to changes in the routine in the future. Better to just let Electron tell us what it’s going to load, which is where `onBeforeRequest` comes in.

It’s also possible to get a failure event before any loading happens. That’s usually because the URL was just totally malformed and fails basic validation. Happily `did-fail-load` is still called in that case, though `webRequest.onErrorOccurred()`—so it’s not feasible to just use webContent events or just use `webRequest` listeners; we have to use them in combination to handle the full range of possible situations.

Whew! I think that covers it; please review for any issues!

Mr0grog:
I think #556 might fix the CircleCI issues here. Not 100% sure, though. If not, I’ll have to poke at it more :\ (Tests run fine for me locally.)

Mr0grog:
Well, still seeing errors here on Circle even with #556 merged in. Oddly, I don’t see them if I SSH into Circle and run them on the console. Not entirely sure what’s up; probably going to give it a rest for tonight.

rosshinkley:
@Mr0grog re your previous comment: like I said, my knowledge there was limited.  Thanks for setting me straight.  :)

I _really_ like `.goto()` returning information about the request.  One question about that, and this may be silly: I see in your tests that [valid redirects are okay](https://github.com/segmentio/nightmare/pull/553/files#diff-910eb6f57886ca16c136101fb1699231R233), but it doesn't check the internal information bubbled up from `.goto()`.  Based on the changes you've put in, would that show the 3xx?  At first blush, I don't think it would - wouldn't it show the 2xx (assuming success) for the final destination (after `did-finish-load`)?

To be clear: I'm not sure `.goto()` should track all of the redirect information from a given request, but I do want to clarify what to expect.  Thoughts?

It looks like atom/electron#5029 has been accepted and as such atom/electron#5013 has been fixed.  I don't know what kind of release schedule Electron is on, but they cut releases often enough that I think I'm willing to wait as it reduces the complexity of your solution (seemingly by quite a bit).  What do you think?

Onto the test issues: I can't get the test suite from your fork to run consistently.  The "should fail properly if request handler is present" test fails intermittently.  It looks like once in a while, `did-fail-load` gets emitted after `did-finish-load`.  (That's the difference in debug output anyway.  I'm not clear on the underlying cause yet.)  Coincidentally, that appears to be one of the problems in build 662.  I don't have time to dig into this right this moment, but as time permits today, I'm going to dig into this and the other CircleCI issues.

Mr0grog:
9cfef7c should mitigate one of the issues I saw repeated on Circle last night: trying to load an unknown protocol seems to hang. It doesn’t do so locally, nor does it if I SSH into the Circle box and manually run the tests from the command line. Not sure what’s different when Circle does its test run, but ¯\_(ツ)_/¯

This doesn’t fix the intermittent "should fail properly if request handler is present” issue you found, @rosshinkley. I noticed that sometimes on Circle, too, but wasn’t able to repro locally and haven’t dug into it yet, so if you’ve got ideas, awesome. 

> It looks like once in a while, did-fail-load gets emitted after did-finish-load. (That's the difference in debug output anyway. I'm not clear on the underlying cause yet.)

*Sigh.* Hope this is not what’s really happening, but that would definitely cause this sort of problem :(

> I see in your tests that valid redirects are okay, but it doesn't check the internal information bubbled up from .goto(). Based on the changes you've put in, would that show the 3xx? At first blush, I don't think it would - wouldn't it show the 2xx (assuming success) for the final destination (after did-finish-load)?

Oh, that’s a good question. I’m pretty sure it would show the 2xx for the final destination, but haven’t tested that case. I’m also not sure I’ve tested the case of invalid redirects.

> I'm not sure .goto() should track all of the redirect information from a given request, but I do want to clarify what to expect.

I actually hadn’t thought about this case *at all!* It would be interesting to report on all the redirects involved, but I hadn’t really planned to. So yes, no reason to expect that.

> I don't know what kind of release schedule Electron is on, but they cut releases often enough that I think I'm willing to wait as it reduces the complexity of your solution (seemingly by quite a bit). What do you think?

Yeah, I’m not sure either, though they do certainly release often. Buuuuuuut gathering the response details also relies on the URL check, so https://github.com/electron/electron/pull/5029 no longer removes the need for the `onBeforeRequest` madness. I’ve just filed https://github.com/electron/electron/issues/5074, which would fix this for us.

rosshinkley:
> 9cfef7c should mitigate one of the issues I saw repeated on Circle last night: trying to load an unknown protocol seems to hang. It doesn’t do so locally, nor does it if I SSH into the Circle box and manually run the tests from the command line. Not sure what’s different when Circle does its test run, but ¯_(ツ)_/¯

Hooray for fixing... but [terrified gasp] at Circle internals.  When I'm done spelunking through Electron's guts, I might have a look at `circle.yaml` and related docs to see if it's a setup problem.  My familiarity with CircleCI is _pretty_ pitiful, so it's worth the exercise.

> I noticed that sometimes on Circle, too, but wasn’t able to repro locally and haven’t dug into it yet, so if you’ve got ideas, awesome.

Working on it.  Again, I want to take a look at the CCI docs, make sure it's not something silly.  For the Nightmare side, more on this in a moment, I want to dump out what I've found so far.

> Sigh. Hope this is not what’s really happening, but that would definitely cause this sort of problem :(

Maybe?  It looks like you were [on to something](https://github.com/electron/electron/issues/5031) with the provisional load failure.  That happens _super_ early in the request cycle,  emitting [`did-fail-load` equally as early](https://github.com/electron/electron/blob/702352804239f58e5abcd0b96dbd748b68ab0278/lib/browser/api/web-contents.js#L150-L156) and omitting that it came from the provisional load.  The handler for the provisional load failure also makes a call to `setImmediate`, which in my understanding (which I will readily admit is probably flawed) could prop up a race condition between `did-fail-load` and the other events.

At any rate, in my sandbox, `did-fail-provisional-load` doesn't get emitted for child parts that fail if I watch it directly. Did you experiment with that event?  If so, why opt against it?

> Oh, that’s a good question. I’m pretty sure it would show the 2xx for the final destination...

Curiosity got the best of me, I gave this a scream.  It does.

> ... but haven’t tested that case.

Explicitly tested that case?  No.  Is it worth adding the checks?  Probably.

>  I’m also not sure I’ve tested the case of invalid redirects.

[This doesn't cover it?](https://github.com/segmentio/nightmare/pull/553/files#diff-910eb6f57886ca16c136101fb1699231R237)

> I actually hadn’t thought about this case at all! It would be interesting to report on all the redirects involved, but I hadn’t really planned to. So yes, no reason to expect that.

It would be neat and make Nightmare more useful as a testing tool, but I'm okay for the moment putting it down as a future enhancement.  Bigger fish to fry, all that. :)

Mr0grog:
> The handler for the provisional load failure also makes a call to setImmediate, which in my understanding (which I will readily admit is probably flawed) could prop up a race condition between did-fail-load and the other events.

AH! I had seen that earlier, but didn’t remember it when you mentioned the race condition. This could definitely cause it. Hmmmm.

> in my sandbox, did-fail-provisional-load doesn't get emitted for child parts that fail if I watch it directly. Did you experiment with that event? If so, why opt against it?

I did look into it, but ultimately just used `did-fail-load` as an expedient for not needing to hook an extra event `did-fail-provisional-load` (because one triggered the other). The potential for a race did not occur to me, though. Just hooking `did-fail-provisional-load` to `handleFailure` would fix things if the `setImmediate` call is in fact the real culprit here.

>> I’m also not sure I’ve tested the case of invalid redirects.
> This doesn't cover it?

I forgot what I wrote! Maybe didn’t get enough sleep last night :P

> [capturing redirect info] would be neat and make Nightmare more useful as a testing tool, but I'm okay for the moment putting it down as a future enhancement.

Agreed 👍 

Side note: I realized while looking at https://circleci.com/gh/Mr0grog/nightmare/13 that I didn’t really do a very good job with the “fail if the response dies in flight” test. Going to rewrite it by adding a server endpoint that just kills the socket, which should make the test reliable.

Mr0grog:
Also, on the `onBeforeRequest` stuff still being needed for the `did-get-response-details` event—I’m happy letting this PR sit while we wait to see what happens with https://github.com/electron/electron/issues/5074 if you want.

rosshinkley:
> AH! I had seen that earlier, but didn’t remember it when you mentioned the race condition. This could definitely cause it. Hmmmm.

It was a bit of a (pleasant) surprise coming across your name.  I figured you had either forgotten or had a good reason for not pursuing it further.  (Lack of sleep is a legitimate reason.)

> The potential for a race did not occur to me, though. Just hooking did-fail-provisional-load to handleFailure would fix things if the setImmediate call is in fact the real culprit here.

For the sake of completeness, it's probably worth trying, if nothing else than to rule it out.  I'll give it a whirl and report back.

> Side note: I realized while looking at https://circleci.com/gh/Mr0grog/nightmare/13 that I didn’t really do a very good job with the “fail if the response dies in flight” test. Going to rewrite it by adding a server endpoint that just kills the socket, which should make the test reliable.

Right on.  Thanks!

> Also, on the onBeforeRequest stuff still being needed for the did-get-response-details event—I’m happy letting this PR sit while we wait to see what happens with electron/electron#5074 if you want.

Yeah, let's let it ride for a bit.  The maintainers there are _generally_ pretty quick with responses.

rosshinkley:
>> The potential for a race did not occur to me, though. Just hooking did-fail-provisional-load to handleFailure would fix things if the setImmediate call is in fact the real culprit here.

> For the sake of completeness, it's probably worth trying, if nothing else than to rule it out. I'll give it a whirl and report back.

I'm still having failures, but I think it's for an entirely different reason.  I'm not sure if it's a resource issue or what, but somewhere in the neighborhood of 50-60 runs, my sandbox (on my low-powered laptop) will hang when starting an Electron process.  (I feel like there's an open issue for this already - #238, maybe?)

Regardless, that's *way* better than the ~1 in 5 failure rate I was seeing before, and the symptoms are totally different judging from the debug info.  From that, I think it's increasingly likely this is a race condition, and as such, swapping out for `did-fail-provisional-load` is a decent call.  Thoughts?

rosshinkley:
Offtopic, but do you (or anyone else) know why `dom-ready` events are emitted even for provisional load failures?

Mr0grog:
> I'm not sure if it's a resource issue or what, but somewhere in the neighborhood of 50-60 runs, my sandbox (on my low-powered laptop) will hang when starting an Electron process. (I feel like there's an open issue for this already - #238, maybe?)

Oh, that issue’s interesting; I’d never seen it. I feel like there are potentially a lot of causes there. Things like #493 could potentially contribute. I also recall seeing some issue about listeners not always being properly cleaned up that didn’t seem like it got well resolved (can’t find now). #556 should reduce the load a little, too.

I also wonder if moving to an architecture where every Nightmare instance is just a new `BrowserWindow` rather than a whole new process (so instead of one Electron process per Nightmare instance, one single Electron process that contains many windows, one for each nightmare instance).

But bottom line, no certainties here, other than that I generally suspect more than one cause. :\

> From that, I think it's increasingly likely this is a race condition, and as such, swapping out for did-fail-provisional-load is a decent call. Thoughts?

Seems reasonable to me. Will implement it as soon as I get a change.

Mr0grog:
I also switched to generating a new in-memory partition for every test to avoid caching issues here; it’s possible that be exacerbating things, too: https://github.com/segmentio/nightmare/pull/553/files#diff-910eb6f57886ca16c136101fb1699231R81

rosshinkley:
Mmmm, I might have jumped the gun in my myopic look at that one test.  Some of the other tests are now failing.  I don't have time right this second to dive into it, but will carve out time tonight.

Mr0grog:
> do you (or anyone else) know why dom-ready events are emitted even for provisional load failures?

This is deep deep down (all the load-related events on WebContents come from way deep in Chromium), but it looks like, after sending the failure notification (https://chromium.googlesource.com/chromium/src/+/master/content/renderer/render_frame_impl.cc#3064), Chromium then loads, as part of the same navigation (so no new request cycle), the error page (https://chromium.googlesource.com/chromium/src/+/master/content/renderer/render_frame_impl.cc#3093). That would, I expect, trigger the `dom-ready` event. Not super straightforward, though.

Mr0grog:
> Some of the other tests are now failing.

By any chance, is that “should fire an event on page load failure?” If so, 52958f219813f883b85b7dcad909d94349b06b42 should fix it.

Mr0grog:
OH MY GOD THE TESTS PASSED ON CIRCLE

rosshinkley:
> Chromium then loads, as part of the same navigation (so no new request cycle), the error page ...

Ahhh, that makes sense.

>> Some of the other tests are now failing.

> By any chance, is that “should fire an event on page load failure?” If so, 52958f2 should fix it.

No, it was one of the `.type()` tests, but that family of tests has been flaky for a _long_ time.  It's spookily similar behavior to this: doesn't fail on decent machines, intermittently fails on modest machines, and seems to almost always fail in CI (at least, the first time).  It's next on my bug hunt docket.

> OH MY GOD THE TESTS PASSED ON CIRCLE

/chuckling

Mr0grog:
Electron PR for `resourceType` on `did-get-reponse-details` that should allow us to drop the funky stuff this PR does with `onBeforeRequest`: https://github.com/electron/electron/pull/5095

Mr0grog:
I just added documentation in the README for this. (Can’t believe I forgot to do that!) I’ve also updated the code so it no longer uses `webRequest` listeners, but will wait to actually push that until there’s an actual electron release that it works with.

rosshinkley:
@Mr0grog Thank you for all of the work you've done for this. :)

Mr0grog:
The race condition with `did-finish-load` and `did-fail-load` will be solved in the next Electron release, too! https://github.com/electron/electron/issues/5128 Wooo upstream fixes making things easier here :)

Mr0grog:
@rosshinkley I think this is good to go. All green and no webRequest stuff.

rosshinkley:
@Mr0grog _Awesome._  I'll put this out as soon as I can carve the time out.

PR #554, opened 2016-04-05
Pass uncaughtExceptions back to the default handler after cleanup

pci:
Nightmare adds an `uncaughtException` listener to clean up the child process. However adding a listener [overrides the default behaviour](https://nodejs.org/docs/latest/api/process.html#process_event_uncaughtexception) and can cause an exit code of zero.

### Steps to reproduce:
```js
// a.js
var Nightmare = require("nightmare");
var n = Nightmare();
var a = require("does-not-exist");
```
And then running `node a.js; echo $?` show the error but reports an exit code of `0`.

### Fix:
Re-throw the error so node can pick it back up. I also changed the `console.log` to a `debug` as the default handler prints a stack trace.

### DIFF
--- a/lib/nightmare.js
+++ b/lib/nightmare.js
@@ -80,9 +80,9 @@ function Nightmare(options) {
 
   process.setMaxListeners(Infinity);
   process.on('uncaughtException', function(err) {
-    debug(err.stack);
+    console.error(err.stack);
     endInstance(self);
-    throw err;
+    process.exit(1);
   });
 
   // if the process nightmare is running in dies, make sure to kill electron
--- a/lib/nightmare.js
+++ b/lib/nightmare.js
@@ -80,8 +80,9 @@ function Nightmare(options) {
 
   process.setMaxListeners(Infinity);
   process.on('uncaughtException', function(err) {
-    console.error(err.stack);
+    debug(err.stack);
     endInstance(self);
+    throw err;
   });
 
   // if the process nightmare is running in dies, make sure to kill electron
--- a/lib/nightmare.js
+++ b/lib/nightmare.js
@@ -82,7 +82,11 @@ function Nightmare(options) {
   process.on('uncaughtException', function(err) {
     console.error(err.stack);
     endInstance(self);
-    process.exit(1);
+    // This allows any user defined 'uncaughtException' handlers
+    // to run before exiting
+    process.nextTick(function() {
+        process.exit(1);
+    });
   });
 
   // if the process nightmare is running in dies, make sure to kill electron
--- a/test/files/nightmare-error.js
+++ b/test/files/nightmare-error.js
@@ -1,0 +1,5 @@
+// This script is used to start nightmare 
+// but then throw a user space error
+var Nightmare = require('../..');
+var nightmare = Nightmare();
+throw new Error("uncaught");
--- a/test/index.js
+++ b/test/index.js
@@ -65,6 +65,16 @@ describe('Nightmare', function () {
     });
   });
 
+  it('should exit with a non-zero code on uncaughtExecption', function(done) {
+    var child = child_process.fork(
+      path.join(__dirname, 'files', 'nightmare-error.js'), [], {silent: true});
+
+      child.once('exit', function(code) {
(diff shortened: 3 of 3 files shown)


Mr0grog:
Given that I’m the person who made this mistake, it would be good if this patch included a test. You could probably just check the exit code in the `exit` event in the test: https://github.com/segmentio/nightmare/blob/master/test/index.js#L48 …or write a similar test that does so if the conditions need to be slightly different.

lib/nightmare.js
```diff
@@ -80,8 +80,9 @@ function Nightmare(options) {
 
   process.setMaxListeners(Infinity);
   process.on('uncaughtException', function(err) {
-    console.error(err.stack);
+    debug(err.stack);
```
Mr0grog: Since this is handling the exception in the same process as a user’s code, shouldn’t we be using `console.error` instead of `debug` so that we don’t swallow stack traces originating from their code? Or am I misreading this?

pci:
Hi @Mr0grog, that's for the feedback. I thought about adding a test, but didn't as you're not meant to do anything after an uncaughtException in user space. However I've looked through some of your tests and seen you have some (rather nice!) tests using child processes so I'll have a go at encapsulating a test.

lib/nightmare.js
pci: I switched to `debug` just because re-throwing the error will bubble through to a node handler which logs the stack trace to `stderr`. Adding `console.error` here means the stack gets printed twice.

The down side to re-throwing is it shows the error originating from here, however rest of the error stack is normal.

pci:
Test added

lib/nightmare.js
Mr0grog: Ah! Totally missed that. I wonder if `process.exit(1)` (I gather `1` is correct based on https://nodejs.org/api/process.html#process_exit_codes) instead of re-throwing would work? Then we’d be in control of logging the correct stack.

pci: Good point, however that means any `process.on('uncaughtException',` that the user has added might not get called. However I've just tested it and neither does re-throwing! Looks like Node handles an error inside an `uncaughtException` handler as a special case. I've switched to `console.error` and `process.exit(1)`.

Mr0grog: > however that means any process.on('uncaughtException', that the user has added might not get called.

Oh, good point. Maybe wrapping the `exit()` call in `nextTick`?

```js
console.error(err.stack);
endInstance(self);
process.nextTick(function() {
  process.exit(1);
});
```

pci: Nice suggestion, feels a little hacky but I can't see much to go wrong as either the user exits in a handler of their own or the script exits on the next tick. Added and also added a comment about why it's being done this way.

Mr0grog:
Thanks for cleaning up my mess! 😅

pci:
@Mr0grog No problem, it was only because we introduced an error into our test framework that we found this bug, so you're in good company! Do you want this rebased into one commit?

Mr0grog:
I'm not a committer, so that's above my pay grade. @matthewmueller’s your man (or maybe @rosshinkley; it seems like he might have the commit bit now and is much more active).

rosshinkley:
I'd leave it as-is, I like having the history, and this isn't all that noisy.  I don't see a real need for a rebase.  (I also don't _really_ know what would be considered "most correct," but that's the stand I'm taking, at least for now.)

I do have commit rights on github and publish rights on npm.  (Hooray!)  I'll be putting together a milestone + release ... soonish.  At a minimum, I'd like to get this + #553 in.  The underlying Electron PRs (thank you, @Mr0grog) have all been accepted, but I don't think they've all been released (at least at the time of this comment).

Mr0grog:
> At a minimum, I'd like to get this + #553 in

If I were going to make a recommendation, I'd say this and #556 are good candidates for a patch release (v2.3.1) since they are small, make minor quality fixes without new features, and are immediately ship-able without other upstream fixes (like #553).

rosshinkley:
@Mr0grog fair enough.  Can always cut another patch when #553 is done.  Will do as time permits.

PR #565, opened 2016-04-12
run tests under an `xvfb` wrapper when headless

### LINKED ISSUE #561: instance churn causes xvfb to occasionally hang
I wasn't sure where to put this - a new issue felt the most logical.

Digging into the test failures that happen intermittently, it looks like the behavior that causes the tests to fail also will present in regular applications (albeit seemingly not as often).  Consider the following (extremely contrived) case:

```js
//create a fake url payload for serial execution
var urls = [];
for (var i = 0; i < 500; i++) {
    urls.push('http://localhost:7600/navigation');
}
//go to each url, and then end
urls.reduce(function(accumulator, url, ix) {
    return accumulator.then(function(results) {
      console.log(ix);
      return (Nightmare())
        .goto(url)
        .end()
        .then(function() {
          console.log('results');
          results.push(ix);
          return results;
        });
    });
  }, Promise.resolve([]))]
  .then(function(results) {
    console.log('done');
    //console.dir(results);
  }).catch(function(){
    console.dir(arguments);
  });          
```

This example creates a _lot_ of instance churn, causing Electron instances to be spun up and killed, much like how the unit test suite works.  After running many times (I've had it happen in as few as 8 and as many as ~450 iterations, seems to be luck of the draw), Nightmare will hang.  In a test context, this will cause Mocha to timeout, which at least partially explains the behavior with the intermittent test failures.

The hang appears to happen when a new `BrowserWindow` instance is created.  Enabling `ELECTRON_ENABLE_LOGGING` and `ELECTRON_ENABLE_STACK_DUMPING` doesn't yield useful information.  Piping Electron output directly to the parent process `std*` buffers also doesn't yield anything terribly interesting.

What I'd like to try next:

1. Try to get `node-inspector` working with Electron.  So far, my attempts to get a debugger attached have been met with limited success.
1. Create `BrowserWindow` churn natively under Electron to see if the behavior can be recreated internally to Electron.
1. If that doesn't hang, I'll take a look at creating browser churn with a simplified node process -> electron process.  That should help narrow things down.
1.  If all else fails, I'll try to make a custom Electron build to try and figure out what's going on.

I'm open to suggestions and ideas on all points.

---

Some other peculiar (possibly unrelated?) things that cropped up as a result:

1. The event forwarding method _always_ throws an error after the `.kill()` call is made because the window is destroyed.  I'll be pulling together a small bugfix for that as time permits.  (I don't think it has any bearing on how the rest of the project behaves, but it certainly can't hurt to fix.)
1. The `close` event case for a clean exit never happens unless `win.close` or `win.destroy` are manually called.  In other words, you'll never see [`electron child process exited with code 0: success!`](https://github.com/segmentio/nightmare/blob/master/lib/nightmare.js#L70-L76).

rosshinkley:
Changes the makefile to run tests under an `xfvb` wrapper called `bb-xvfb` under one or more the following conditions:
 * `make test` is executed on CircleCI.
 * `make test` is executed on a box where `xvfb` is running.
 * `make test` is executed in an environment where the `HEADLESS` variable is set and not equal to `1`.

(Disclaimer: my ability with GNU Make is pitiful, any pointers on how to make the Makefile better would be appreciated.)

The wrapper enforces starting `xvfb`, waiting for an X server to be ready, followed by starting `dbus` for IPC messaging, followed by your command.  In the case of `make test`, the unit tests are executed.

The original problem is explained in [Chromium bug 309093](https://bugs.chromium.org/p/chromium/issues/detail?id=309093), and I found that the [Protractor](http://angular.github.io/protractor/#/) project (an Angular test automation suite that uses a Chrome WebDriver to run tests) already had hit this problem.  The scripts used are ~~plagiarised~~ borrowed from [angular/protractor#2419](https://github.com/angular/protractor/issues/2419#issuecomment-156527809).  (Thanks to @tullmann for providing the fix.)

This PR should fix #561.

Thoughts?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,8 +1,26 @@
 GREP ?=.
+#check if xvfb is running
+XVFB_RUNNING = $(shell pgrep "Xvfb" > /dev/null; echo $$?)
+#set the circle project name if not set
+CIRCLE_PROJECT_REPONAME ?= 1
+#set headless if not set
+HEADLESS ?= 1
+
+#debug the environment variables
+$(info $(CIRCLE_PROJECT_REPONAME)$(HEADLESS)$(XVFB_RUNNING))
 
 test: node_modules
+#if this build is not on circle, is not headless, and xvfb is not already running,
+#run mocha as usual
+#otherwise, run mocha under the xvfb wrapper
+ifeq ($(CIRCLE_PROJECT_REPONAME)$(HEADLESS)$(XVFB_RUNNING), 111)
 	@rm -rf /tmp/nightmare
 	@node_modules/.bin/mocha --harmony --grep "$(GREP)"
+else
+	echo 'running under xvfb wrapper'
+	@rm -rf /tmp/nightmare
+	@./test/bb-xvfb node_modules/.bin/mocha --harmony --grep "$(GREP)"
+endif
 
 node_modules: package.json
 	@npm install
--- a/Makefile
+++ b/Makefile
@@ -17,7 +17,6 @@ ifeq ($(CIRCLE_PROJECT_REPONAME)$(HEADLESS)$(XVFB_RUNNING), 111)
 	@rm -rf /tmp/nightmare
 	@node_modules/.bin/mocha --harmony --grep "$(GREP)"
 else
-	echo 'running under xvfb wrapper'
 	@rm -rf /tmp/nightmare
 	@./test/bb-xvfb node_modules/.bin/mocha --harmony --grep "$(GREP)"
 endif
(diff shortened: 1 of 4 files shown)


Mr0grog:
Extra minor note here (sorry I completely missed you posting this PR, @rosshinkley): it might be nice if the two new scripts got sequestered into their own directory since they aren’t really part of the tests per se. Maybe `/scripts` or `/test/scripts`?

rosshinkley:
@Mr0grog Ah, that's no problem.  I pulled it in because I was getting frustrated with documentation-only PRs failing in CircleCI.  Irritation-driven acceptance is a thing, right?

Kidding aside, that's not a bad call.  I'll put it on the list. :)

Mr0grog:
> Irritation-driven acceptance is a thing, right?

Not until you write a post about it on Medium! ;)

PR #573, opened 2016-04-15
.evaluate waits for promise to settle in document

### LINKED ISSUE #552: Wait for promise to settle in document
It would be great if there was a means to `.evaluate()` and wait on a promise. For example:

```JS
nm.evaluate(function(nmjsResolver, nmjsRejector){
   navigator.api.
    .waitForThing
    .then(result => nmjsResolver(result))
    .catch(nmjsRejector);
});
```

This could avoid the unnecessary polling that Nightmarejs currently does while waiting for values to become true in the document.

lzrski:
If a function passed to `.evaluate` return a promise then nightmare will wait for it to be resolved before sending the value (or reject before sending an error). This allows to execute async tasks in `.evaluate` and closes #552.

Also template strings are used for javascript templates (both `execute` and `inject`) - closes #447.


lzrski:
Hmm... Failing tests in CI doesn't seem to be related to the change, esp. since I have only added two test cases and before that CI was passing. I had some troubles with screenshots related tests before. Is it possible they are not deterministic?

rosshinkley:
This change affects `.wait()` as well as they share the same template.  Would it be worth adding a `.wait()` test for promises as well?

As for the failing tests, I issued a rebuild and it appears to have succeeded.  I think your PR was after 2.3.2 was released, so the `xvfb` fix _should_ have been included.  Beyond that, I'm not sure why the tests are failing, might be worth another look if it happens again.

lzrski:
Thanks for your feedback. I'll be happy to do it.

What do you think should be expected behaviour of wait when promise is returned? I guess Nightmare should wait for it to be settled.

What if it's rejected?

Also, if it's fulfilled, should we care about value?

rosshinkley:
@lzrski Off the cuff, I'd say have the Nightmare process to wait for the promise to be settled, yes.

If it's rejected, I'd think it should handle it like `.evaluate()`: bubble the rejection all the way up.  A failure is a failure.  Thoughts?

As for the fulfillment value, I think it should work like how `.wait(fn)` works - the waited-for value should resolve to `true` to stop waiting.  Fortunately, you've already written the plumbing to do this.  All you _should_ need to do is write the test.

lzrski:
I looked at existing test cases for `wait` and I think we have a problem here. They are not really testing anything. There are no assertions there. As long as the code doesn't throw the tests are passing even if behaviour is wrong.

Further more following test case executed in current `master` passes immediately instead of timing out:

```javascript
it.only('should wait until the evaluate fn returns true', function*() {
  yield nightmare
    .goto(fixture('navigation'))
    .wait(function () {
      return false;
      // var text = document.querySelector('a').textContent;
      // return (text === 'A');
    });
});
```

Am I right this proves a bug in `wait`? If you agree I will file an issue.

rosshinkley:
> I looked at existing test cases for wait and I think we have a problem here. They are not really testing anything. There are no assertions there. 

What would be asserted?  `.wait()` doesn't have a return/resolution value (what would it return?), so there's nothing really check against.  The existing tests verify that wait eventually returns and doesn't except.  In that regard, they're safeguarding from errant behavior.  The tests _are_ missing expected failure tests, though.  I can add that to my todo if you'd like. 

> Further more following test case executed in current master passes immediately instead of timing out:

Hm, this appears to have something to do with `it.only`, and I suspect it's a problem in `mocha-generators`.  On my local instance leaving it with `it`, the 10s `waitTimeout` default is hit.

rosshinkley:
`it.only` problem with mocha genators is confirmed.  See vdemedes/mocha-generators/pull/3.

lzrski:
You are right about `.only` being a problem. I got expected results (timeout) without it as well. I discovered that you can set `.only` on describe to get reasonably fast test cycle without having this side effect. It is certainly a bug in mocha / mocha-generators though.

What could be asserted is the passing of time. Currently this case looks more like it is testing if element `a` will eventually get text content of `'A'`. The correctness of `.wait` is kind of assumed. If at some point a bug is introduced and `.wait` returns immediately no matter what the return value of callback is, the tests would not discover it.

How about something like this:

```javascript
it('should wait until the evaluate fn returns true', function*() {
  const beggening = Date.now();

  yield nightmare
    .goto(fixture('navigation'))
    .evaluate(function() {
      window.now_is_the_time = false;
      setTimeout(function() {
        window.now_is_the_time = true;
      }, 1000);
    })
    .wait(function () {
      return window.now_is_the_time;
    });

  beggening.should.be.below(Date.now() - 1000);
});
```

rosshinkley:
I'm not against that sort of test.  I was thinking more on the lines of rejection testing, which would also catch the problem you've outlined.  Off-the-cuff, something like:

```js
if('should error on timeout', function(){
  var nightmare = Nightmare({waitTimeout:500});
  return nightmare.goto(fixture('navigation'))
    .wait(function(){return false;})
    .should.be.rejected;
})
```

test/index.js
```diff
@@ -160,6 +160,20 @@ describe('Nightmare', function () {
           return (expectedA === textA && expectedB === textB);
         }, 'A', 'B');
     });
+
+    it('should wait for returned promise to be resolved', function*() {
+      const beggening = Date.now();
```
rosshinkley: Super-minor spelling nit: `beggening` => `beginning`

lzrski:
Oh, it's a good idea to test for falsy resolution of a promise! Let me push again :)

test/index.js
lzrski: Thanks :P

rosshinkley:
:) I'd say both are reasonable additions.

rosshinkley:
Rereading your submission, I spotted something I'm shocked I haven't (and no one else has) spotted.  Having `.wait()` support asynchronous methods _might_ not work as expected.  Consider something like:

```js
nightmare
  .goto(someUrl)
  .wait(function(){
    return new Promise(function(resolve){
      //what if promise internals take a while to process?
      setTimeout(resolve, 60000);
    });
  });
```
... would not cause a timeout as the current version of Nightmare to timeout as written.  Wouldn't that cause `waitTimeout`/250 (the tick length) iterations of that promise to run instead of timing out after `.waitTimeout` waiting for the promise?  In fact, now `.wait()` won't time out until (`waitTimeout`/250)*execution time + `waitTimeout`.  This is not good.

Some off-the-top-of-my-head solutions:

* Some sort of watcher method that tracks time elapsed instead of tracking it internal to `.wait()`
* Split the templates for `.wait()` and `.evaluate()`, pass the timeout down to `.wait()`, and have the template source send back an error message if the timeout is hit.
* Leave it as-is.  I'm worrying too much.

Thoughts?

lzrski:
Yes, when you run test like that:

```javascript
it('should wait for returned promise to be resolved to truthy value', function() {
  var nightmare = Nightmare({waitTimeout:1000});

  return nightmare
    .goto(fixture('navigation'))
    .wait(function() {
      return new Promise(function(resolve) {
        setTimeout(resolve, 2000, false);
      })
    })
    .then(function () {
      throw Error("Promise should have been rejected.");
    })
    .catch(function(error) {
      error.message.should.be.equal(".wait() timed out after 1000msec");
    });
})
```

then Nightmare's timeout is never enforced. It's Mocha that times out after 10s. I think it's a legitimate issue and we should deal with it.

Splitting the templates doesn't seem compelling. Shouldn't we have a timeout for `.evaluate` just as well as for `.wait`? It would probably prevent some hard to debug errors in applications using Nightmare. So I would go with your first idea. Just not sure where this logic would fit.

lzrski:
It seems the timeout logic is in `waitFn` function in `actions.js`. It seems quite complex. Couldn't simple `setTimeout` be used?

rosshinkley:
It occurs to me that this problem exists now.  Consider:

```js
nightmare
  .goto(url)
  .wait(function(){
    while(true){ }
  })
  ...
```
... or any long-running synchronous loop.  Same problem, slightly different cause.

> Splitting the templates doesn't seem compelling. 

I've thought about this a bit more, and realized that pushing the timeout responsibility to the evaluated client script is not a good idea as then the timeout is at the mercy of the client page.  This could lead unintended behavior if the client page clears timeouts.  (I was shooting from the hip.)

> Shouldn't we have a timeout for .evaluate just as well as for .wait?

An evaluation timeout seems like a reasonable addition, so long as the default behavior is the current behavior.

> Just not sure where this logic would fit.

I'll have to think about it, but I _think_ the most logical spot would be actually the `javascript` listener in `runner.js`.  That buys both `.wait()` and `.evaluate()` timeouts.  I'd have to play with the source to determine feasibility, though.

> Couldn't simple setTimeout be used?

Depends on what you mean.  The `.wait(fn)` and I believe `.wait(selector)` methods poll the client page every 250ms to determine if something has changed (eg, something is added, or a DOM element changes color).  I don't think you're proposing simply waiting `waitTimeout` and checking at the end, but if so, that would not be a good idea.

If you are proposing having a `setTimeout` handle the error case where `waitTimeout` has elapsed (and I believe you are, correct me if I'm wrong), that's a good idea.  That may be the best approach, it's then a matter of where to implement it.

carebdayrvis:
Anything I can do to help this along?

lzrski:
Hey, sorry for abandoning this conversation. It's because my first child was recently born and I was quite occupied with this wonderful if somewhat difficult change in my life :)

@carebdayrvis thanks for asking. Any help is appreciated.

Back to the topic. It seams that `waitFn()` calculates the time elapsed by just adding 250ms after every tick. This way it ignores the time that the evaluation on page took. With async evaluation this can be significant. Am I understanding correctly?

This could be improved by storing the start timestamp and comparing it with current time on every tick, like that:

```javascript
var _timeoutMs = 250;
function waitfn() {
  var startTime = Date.now();
  // instead of var waitMsPassed = 0;
  return tick.apply(this, arguments)

  function tick (self, fn/**, arg1, arg2..., done**/) {
    var args = sliced(arguments);
    var done = args[args.length-1];
    {+var timeElapsed = Date.now() - startTime;+}
    var waitDone = function (err, result) {
      if (result) {
        return done();
      }
      else if (self.optionWaitTimeout && [-waitMsPassed-]{+timeElapsed+} > self.optionWaitTimeout) {
        return done(new Error('.wait() timed out after '+self.optionWaitTimeout+'msec'));
      }
      else {
[-        waitMsPassed += _timeoutMs;-]
        setTimeout(function () {
          tick.apply(self, args);
        }, _timeoutMs);
```

but still wouldn't work for very long running evaluations - it will only timeout after the job is done, and only if the result if negative (`false` is returned). The commit that I've just pushed demonstrates this by introducing a (failing) test case.

lzrski:
Whoa! Everything is timing out in the CI.

🤔

lzrski:
Ok, merged master into this PR's branch and it **fails as expected** now. Guess the configuration of the CI environment changed and some adjustments were needed in the code base as well.

What's the status of #609? It looks like our efforts may collide.

rosshinkley:
Reimplented and fixed in #819.

Issue #572, opened 2016-04-15
Test for devtools seem to fail inderterministically


lzrski:
Hello, I'm playing with Nightmare and mostly love the experience. Thanks for this tool.

It seems like electron instance is sometimes (but not always) crushing when openDevTools option is set to `true`. I was not able to find a pattern. Here is a debug log for failing case:

```
 $ DEBUG=nightmare* npm test

> nightmare@2.3.2 test /Users/tadeuszlazurski/Projects/nightmare
> make test



  Nightmare
    devtools
  nightmare queueing child action addition for "checkDevTools" +0ms
  nightmare queueing action "goto" for http://localhost:7500/simple +5ms
  nightmare queueing action "wait" +1ms
  nightmare queueing action "checkDevTools" +0ms
  nightmare running +0ms
  nightmare:log adding action for checkDevTools +289ms
  nightmare:log did-start-loading +57ms [{"sender":{"session":{"cookies":{},"webRequest":{}},"hostWebContents":null,"devToolsWebContents":null,"debugger":{},"_maxListeners":0,"_events":{"page-title-updated":[null,null],"crashed":[null,null]},"_eventsCount":28,"browserWindowOptions":{"show":true,"openDevTools":true,"alwaysOnTop":true,"webPreferences":{"preload":"/Users/tadeuszlazurski/Projects/nightmare/lib/preload.js","nodeIntegration":false}}}}]
  nightmare:log did-get-redirect-request +38ms [{"sender":{"session":{"cookies":{},"webRequest":{}},"hostWebContents":null,"devToolsWebContents":null,"debugger":{},"_maxListeners":0,"_events":{"page-title-updated":[null,null],"crashed":[null,null],"did-finish-load":[null,null]},"_eventsCount":27,"browserWindowOptions":{"show":true,"openDevTools":true,"alwaysOnTop":true,"webPreferences":{"preload":"/Users/tadeuszlazurski/Projects/nightmare/lib/preload.js","nodeIntegration":false}}}},"http://localhost:7500/simple","http://localhost:7500/simple/",true,303,"GET","",{"connection":["keep-alive"],"content-length":["47"],"content-type":["text/html; charset=UTF-8"],"date":["Fri, 15 Apr 2016 08:34:10 GMT"],"location":["/simple/"],"x-content-type-options":["nosniff"],"x-powered-by":["Express"]}]
[...]

Issue #563, opened 2016-04-12
openDevTools hangs the electron instance.


itsyogesh:
Hi there,

I was trying to get openDevTools to work for debugging, but as soon as it opens the electron instance, it hangs the operation and exits after a few seconds. Is this an issue, since the openDevTools was added recently and I couldn't find any tests to check if I am doing this correctly.

Here is the debug output:
```
nightmare:log did-start-loading +508ms [{"sender":{"session":{"cookies":{},"webRequest":{}},"hostWebContents":null,"devToolsWebContents":null,"debugger":{},"_maxListeners":0,"_events":{"page-title-updated":[null,null],"crashed":[null,null]},"_eventsCount":28,"browserWindowOptions":{"show":true,"openDevTools":true,"alwaysOnTop":true,"webPreferences":{"preload":"/Users/Yogesh/Projects/drift/node_modules/nightmare/lib/preload.js","nodeIntegration":false}}}}]
  nightmare electron child process exited with code null: undefined +4s
```

rosshinkley:
Hm, that's one of mine, and I should have added unit tests for it.  I'll add that to my list.

What bothers me about your particular case is the "exited with code null" bit: The only time that happens (at least that I'm aware of) is when a segmentation fault occurs in Electron.  Could you provide a minimum case that reproduces the issue?

rosshinkley:
Suspicion confirmed, this is a segfault internal to Electron that appears to happen with the version of Electron Nightmare is currently on.  It does not happen with the current version of `electron-prebuilt` (0.37.5 at the time of this comment).  Curiously, electron/electron#4103 is the closest problem I can find in the repository.  It's not exactly the same thing, but the symptoms are very similar.

At any rate, I'll put together a PR with the requested tests and a version bump.

lunow:
In the latest version, `webPreferences` setting on nightmare object causing this problem for me: electron instance is crashing immediately.

rosshinkley:
@lunow if that's true, it's likely got a different cause.  Could you open a new issue with a minimal reproduction case?

PR #576, opened 2016-04-15
Workaround for Electron event sender crash

Mr0grog:
Electron has a potential crasher (on OS X, at least) when the devtools are open: https://github.com/electron/electron/issues/5180

This is an extremely simple workaround. Nightmare’s event forwarding code sends the `event.sender` argument, which may contain properties that crash Electron when accessed. Nightmare triggers the crash by stringifying everything when forwarding the event across IPC. This simply avoid including the `sender`.

I *think* it would be nicer not to send the `event` object at all (since this patch leaves it as an empty object in most cases and I don’t think it provided any users with useful data), but I am ever so slightly worried that changing which arguments are where in the events could bite somebody. This seemed safer.

### DIFF
--- a/lib/runner.js
+++ b/lib/runner.js
@@ -356,9 +356,9 @@ app.on('ready', function() {
 
   parent.on('action', function(name, fntext){
     var fn = new Function('with(this){ parent.emit("log", "adding action for '+ name +'"); return ' + fntext + '}')
-      .call({ 
-        require: require, 
-        parent: parent 
+      .call({
+        require: require,
+        parent: parent
       });
     fn(name, options, parent, win, renderer, function(err){
       parent.emit('action', err);
@@ -407,8 +407,12 @@ app.on('ready', function() {
  * Forward events
  */
 
-function forward(event) {
-  return function () {
-    parent.emit.apply(parent, [event].concat(sliced(arguments)));
+function forward(name) {
+  return function (event) {
+    // trying to send the event's `sender` can crash electron, so strip it
+    // https://github.com/electron/electron/issues/5180
+    var safeEvent = Object.assign({}, event);
+    delete safeEvent.sender;
+    parent.emit.apply(parent, [name, safeEvent].concat(sliced(arguments, 1)));
   };
 }
--- a/test/index.js
+++ b/test/index.js
@@ -1188,8 +1188,7 @@ describe('Nightmare', function () {
     });
 
     it('should support extending nightmare', function*() {
-      var nightmare = Nightmare()
-      var tagName = yield Nightmare()
+      var tagName = yield nightmare
         .goto(fixture('simple'))
         .use(select('h1'))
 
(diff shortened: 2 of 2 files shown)


Mr0grog:
Fixes #572, #563.

Mr0grog:
Well, I guess the sped up tests go so fast they complete before Electron has a chance to crash! They’re still valid, but no longer a good canary for the crash this PR is mainly about preventing. Should I roll them back?

rosshinkley:
> Electron has a potential crasher (on OS X, at least) when the devtools are open:

It's happening on my Debian-based Linux boxes too.  I should also mention #578 - I think it's symptoms of the same problem.

> They’re still valid, but no longer a good canary for the crash this PR is mainly about preventing. Should I roll them back?

Uh, I guess so.  Is there a reason to keep the changes?

Mr0grog:
> Is there a reason to keep the changes?

Well, they allow the test to complete in a fraction of a second rather than several seconds (and still have accurate results). Nightmare’s test suite isn’t so big that it’s a problem now, but I’ve seen plenty of projects where that is a problem, so it seemed like a reasonable change while I was poking at the dev tools stuff.

The question, I guess, is whether it’s useful to have a much slower test that *might* also tell us we avoided the crasher (since the crash is intermittent).

rosshinkley:
The crash test isn't causal, making the argument for leaving it as-is kind of weak.  Plus, the speed benefit is nice.  But not having _any_ way of catching the crash also seems silly.  I wish there were a better way of explicitly causing the crash.

I am still on the fence, but tipped toward inclusion.  This problem seems important enough to merit _reasonably_ quick attention from the folks at Electron, so I imagine it will be addressed fairly quickly.  Thoughts?

Mr0grog:
> This problem seems important enough to merit reasonably quick attention from the folks at Electron, so I imagine it will be addressed fairly quickly.

Yeah, I’d imagine so. I would tend towards not worrying about it under that expectation, but we could leave the tests as-is (that is roll back the changes here) until the bug is addressed, and then add the improvements here.

If we really want to test for the crash, we should have an explicit test for it and, since it’s intermittent, probably write it like [the screenshot reliability tests](https://github.com/segmentio/nightmare/blob/master/test/index.js#L717), where it gets repeated a few times to get a better guarantee that everything is working.

Mr0grog:
The crash is fixed upstream in Electron: https://github.com/electron/electron/pull/5191. We’ll have to watch for the next release and update.

Based on that, I think we should just go with the faster test. I also still think cutting the `sender` out of forwarded events would be good since it reduces a lot of noise in the debug logs—once you get to the parent process, it’s no longer a meaningful reference and there’s not much useful data that gets serialized there, either.

rosshinkley:
That's good.  Great, even.  Faster tests make for happy tests.

As for stripping out the sender, I'd be pretty okay with that.  I've never found it _particularly_ useful outside of having the `browserWindowOptions` barfed out with every event, and those could be sent in a one-time debug message.  And I'm not sure _that_ is even necessary.

Issue #421, opened 2015-12-29
swallowing electron errors


stephenmathieson:
we're currently swallowing errors, making debugging nightmare timeouts incredibly difficult:

**timeout.js**

```js

var ms = require('ms');
var Nightmare = require('./')

var n = new Nightmare;
n.goto('http://yahoo.com').end(function() {
  console.log('=)')
  process.exit(0)
});

setTimeout(function() {
  console.error('timeout')
  process.exit(1)
}, ms('30s'))
```

**output**:

```
$ DEBUG=nightmare* node timeout
  nightmare queueing action "goto" for http://yahoo.com +17ms
  nightmare running +3ms
timeout
```

----

running `dist/electron` directly yields:

```
$ node_modules/electron-prebuilt/dist/electron lib/runner.js 
node_modules/electron-prebuilt/dist/electron: error while loading shared libraries: libgconf-2.so.4: cannot open shared object file: No such file or directory
```

... any reason we're not sending this to `stderr`, or at least notifying the user that something went wrong?

shauvik:
I'm trying to run electron on GCP and get a similar error:

```
$ node_modules/electron-prebuilt/dist/electron lib/runner.js 
electron: error while loading shared libraries: libnotify.so.4: cannot open shared object file: No such file or directory
```

Mr0grog:
Since this was filed, it looks like Nightmare has gotten a *little* better at surfacing electron errors: https://github.com/segmentio/nightmare/blob/master/lib/nightmare.js#L70

…but that really wouldn’t give nearly as much detail as it could. A user would never see the message above from electron, for example:

```
node_modules/electron-prebuilt/dist/electron: error while loading shared libraries: libnotify.so.4: cannot open shared object file: No such file or directory
```

One option would be to just send Electron’s stderr through the parent process’s stderr. Users would see deprecation notices and possibly a number of other bits of noisy output from Electron. There’s probably some good to be had there (in trying this, I noticed Nightmare uses several deprecated APIs), but I can imagine there might be other output that is more noisy than helpful.

Another option might be to buffer up Electron’s stderr and emit it when/if Electron exits with a non-zero exit code. Any thoughts?

rosshinkley:
_Bringing this over from #582._

[...]

PR #581, opened 2016-04-18
Fix usage of deprecated Electron APIs

Mr0grog:
We were using a few deprecated Electron APIs in various spots. Best to fix that before they disappear entirely and break! Happily, they were all just renamings or moves to other locations, so this was very straightforward.

Noticed while looking at #421.


rosshinkley:
Looks good.  Thanks!

PR #582, opened 2016-04-18
Add testing for deprecated API usage in Electron

Mr0grog:
Running across #581 made me think it would be a good idea to get automatic warning of deprecated API usage. This causes the test suite to fail at the end if any deprecated APIs are used.

The implementation basically wraps the Nightmare constructor used in tests and watches Electron’s `stderr` for deprecation warnings, then throws an exception at the end if there were any.

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,6 +40,7 @@
     "mocha-generators": "^1.2.0",
     "multer": "1.1.0",
     "pngjs": "^2.2.0",
-    "serve-static": "^1.10.0"
+    "serve-static": "^1.10.0",
+    "split": "^1.0.0"
   }
 }
--- a/test/index.js
+++ b/test/index.js
@@ -15,6 +15,7 @@ var rimraf = require('rimraf');
 var child_process = require('child_process');
 var PNG = require('pngjs').PNG;
 var should = chai.should();
+var split = require('split');
 
 /**
  * Temporary directory
@@ -37,6 +38,11 @@ var base = 'http://localhost:7500/';
 describe('Nightmare', function () {
   before(function (done) {
     server.listen(7500, done);
+    Nightmare = withDeprecationTracking(Nightmare);
+  });
+
+  after(function() {
+    Nightmare.assertNoDeprecations();
   });
 
   it('should be constructable', function*() {
@@ -1245,7 +1251,7 @@ describe('Nightmare', function () {
           this.child.emit('checkDevTools');
         });
       nightmare = Nightmare({show:true, openDevTools:true});
-      
+
     });
 
     afterEach(function*(){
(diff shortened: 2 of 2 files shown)


Mr0grog:
Note the tests *should* fail, since we are using some deprecated APIs :P

#581 should be merged first in order to keep these tests happy.

rosshinkley:
This looks great.

This went in a different direction than I had thought, though.  Not sure if this is the best place for this conversation, but here goes: something I had been thinking about was wrapping the Electron instance stdout and possibly stderr up with calls to `debug` (rather than piping them straight through to the parent process stdio).  Then I started thinking about possibly having log levels and _maybe_ introducing something slightly more complex than debug like [winston](https://github.com/winstonjs/winston).  Thoughts?

Mr0grog:
> something I had been thinking about was wrapping the Electron instance stdout and possibly stderr up with calls to debug (rather than piping them straight through to the parent process stdio). Then I started thinking about possibly having log levels and maybe introducing something slightly more complex than debug like winston.

I generally like that idea, but the place for it is probably #421.

As for *this* PR, I thought it was valuable to programmatically examine stderr for these warnings so we could explicitly cause tests to fail (and if that doesn’t feel great, cool, we shouldn’t take this). We couldn’t do that with just a logger. It also lets us a) group all the warnings together so they aren’t spread though-out the test output and b) dedupe them, so you only see “printToPDF is deprecated.” once, for example:

```
  82 passing (1m)
  1 failing

  1) Nightmare "after all" hook:
     Error: Used 2 deprecated Electron APIs:
        (electron) printToPDF is deprecated. Use webContents.printToPDF instead.
        (electron) use-content-size is deprecated. Use useContentSize instead.
```

Since Electron’s stdio are just streams, there’s no reason we can’t pipe them to multiple places if we want to. Doing this shouldn’t get in the way of piping to `debug` or some other logging tool.

rosshinkley:
> I generally like that idea, but the place for it is probably #421.

Good point, I'll move it there.

> As for this PR, I thought it was valuable to programmatically examine stderr for these warnings so we could explicitly cause tests to fail...

I'm realizing now I might have been unclear.  I completely agree, evaluating stderr output for deprecation is a good idea.  I like having them collected and deduped for the end.  And ...

> Since Electron’s stdio are just streams, there’s no reason we can’t pipe them to multiple places if we want to. Doing this shouldn’t get in the way of piping to debug or some other logging tool.

... was kind of what I was driving at: we can get both, and it might be worthwhile to do so.  I just blanked on #421.

Mr0grog:
Oh, ha, sorry. Violent agreement!

rosshinkley:
Totally fine. :)

PR #584, opened 2016-04-19
piping electron stdout and stderr to debug calls

### LINKED ISSUE #421: swallowing electron errors

rosshinkley:
Fixes #421, at least in part.

Open questions: 

1. Should Nightmare override the environment variable for `ELECTRON_ENABLE_LOGGING`?
2. Should the `nightmare:` namespace be on the debug calls for Electron?

### DIFF
--- a/lib/nightmare.js
+++ b/lib/nightmare.js
@@ -5,8 +5,8 @@
 var log = require('debug')('nightmare:log');
 var debug = require('debug')('nightmare');
 var electronLog = {
-  stdout: require('debug')('nightmare:electron:stdout'),
-  stderr: require('debug')('nightmare:electron:stderr')
+  stdout: require('debug')('electron:stdout'),
+  stderr: require('debug')('electron:stderr')
 };
 
 /**
--- a/lib/nightmare.js
+++ b/lib/nightmare.js
@@ -21,6 +21,7 @@ var join = require('path').join;
 var sliced = require('sliced');
 var child = require('./ipc');
 var once = require('once');
+var split2 = require('split2');
 var noop = function() {};
 var keys = Object.keys;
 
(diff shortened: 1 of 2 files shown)


lib/nightmare.js
```diff
@@ -67,6 +71,14 @@ function Nightmare(options) {
     stdio: [null, null, null, 'ipc']
   });
 
+  this.proc.stdout.on('data', (data) => {
```
Mr0grog: It’s unlikely to be too big a deal, but as a general practice, it’d be a good idea to pipe through `split` or `split2` first—there’s no guarantee otherwise that stuff will come out of the `stdout/stderr` streams as whole lines (could be more or less).

Mr0grog:
> Should Nightmare override the environment variable for ELECTRON_ENABLE_LOGGING

I would say nah. For high-level issues someone is likely to care about (deprecation notices, info about why Electron failed to start, etc.), info gets logged regardless of this env var. It gets pretty verbose if you turn it on. Better to leave that up to a user, I think.

> Should the nightmare: namespace be on the debug calls for Electron?

I’d say nah to this, too, but I don’t really have any good reasoning or a strong case for it.

lib/nightmare.js
rosshinkley: Ah, that's a good call.  I'll make the update.

rosshinkley:
> It gets pretty verbose if you turn it on. Better to leave that up to a user, I think.

My thoughts too.  Thought I'd float it just to make sure there wasn't something I missed.

> I’d say nah to this, too, but I don’t really have any good reasoning or a strong case for it.

The best (and possibly only) reason I'd be in favor of leaving it prefixed with `nightmare:` is to maybe simplify `DEBUG`.  However, with your test to catch deprecation warnings, I'm not sure there's a case to be made.

PR #585, opened 2016-04-20
test nonexistent selectors for `.type()` and `.insert()`, change exceptions to use `chai-as-promised`

### LINKED ISSUE #577: `.type()` does not fail on a non-existent selector
Consider:

```js
nightmare
  .goto(url)
  .type('selector-existing', 'hello')
  .type('selector-not-existing', 'world')
```

`world` would be typed into the already-focused `selector-existing`.

rosshinkley:
Fixes #577.

### DIFF
--- a/lib/actions.js
+++ b/lib/actions.js
@@ -145,7 +145,10 @@ exports.type = function() {
   } else {
     this.evaluate_now(function (selector) {
       document.querySelector(selector).focus();
-    }, function() {
+    }, function(err) {
+      if(err){
+        return done(new Error(`cannot type into selector "${selector}".`));
+      }
       child.once('type', done);
       child.emit('type', text);
     }, selector);
@@ -177,7 +180,10 @@ exports.insert = function (selector, text, done) {
   } else {
     this.evaluate_now(function (selector) {
       document.querySelector(selector).focus();
-    }, function() {
+    }, function(err) {
+      if(err){
+        return done(new Error(`cannot type into selector "${selector}".`));
+      }
       child.once('insert', done);
       child.emit('insert', text);
     }, selector);
--- a/lib/actions.js
+++ b/lib/actions.js
@@ -147,7 +147,7 @@ exports.type = function() {
       document.querySelector(selector).focus();
     }, function(err) {
       if(err){
-        return done(new Error(`cannot type into selector "${selector}".`));
+        return done(new Error(`.type() failed: selector "${selector}" does not exist.`));
       }
       child.once('type', done);
       child.emit('type', text);
@@ -182,7 +182,7 @@ exports.insert = function (selector, text, done) {
       document.querySelector(selector).focus();
     }, function(err) {
       if(err){
-        return done(new Error(`cannot insert into selector "${selector}".`));
+        return done(new Error(`.insert() failed: selector "${selector}" does not exist.`));
       }
       child.once('insert', done);
       child.emit('insert', text);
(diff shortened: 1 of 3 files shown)


lib/actions.js
```diff
@@ -145,7 +145,10 @@ exports.type = function() {
   } else {
     this.evaluate_now(function (selector) {
       document.querySelector(selector).focus();
-    }, function() {
+    }, function(err) {
+      if(err){
+        return done(new Error(`cannot type into selector "${selector}".`));
```
Mr0grog: I think it might be better to have a more explicit message—“can’t find element matching selector” or something along those lines. I can see myself puzzling over this message—“well *why* couldn’t it type?”

Could maybe prefix it with the action name or add a property to the error indicating what action it came from if that’s a concern.

lib/actions.js
```diff
@@ -177,7 +180,10 @@ exports.insert = function (selector, text, done) {
   } else {
     this.evaluate_now(function (selector) {
       document.querySelector(selector).focus();
-    }, function() {
+    }, function(err) {
+      if(err){
+        return done(new Error(`cannot insert into selector "${selector}".`));
```
Mr0grog: Same as above w/r/t error messages.

test/index.js
```diff
@@ -547,6 +547,30 @@ describe('Nightmare', function () {
...
+    it('should not type in a nonexistent selector', function*(){
+      var failed = false;
+      try {
+        yield nightmare
+          .goto(fixture('manipulation'))
+          .type('does-not-exist', 'nightmare');
+      } catch(e){
+        failed = true;
+      }
+      failed.should.be.true;
```
Mr0grog: I wonder if using [Chai-as-promised](http://chaijs.com/plugins/chai-as-promised/) might be a good a way to clean these up. A bunch of the `goto` tests have similar reverse-assertion logic, and could be more easily written:

```js
it('should not type in a nonexistent selector', function() {
  return nightmare
    .goto(fixture(‘manipulation’))
    .type(‘does-not-exist’, ‘nightmare’)
    .then().should.be.rejected;
});
```

lib/actions.js
rosshinkley: Fair.  I'll make the change.

test/index.js
rosshinkley: Mmm, I had not considered this.  I'd be willing to fold it in, but it feels a _bit_ out of scope for this PR... thoughts?

rosshinkley: .... out of scope if you were intent on changing any method that could be cleaned up anyway.  Rereading this, I think you may have intended changing only the rejection tests.

Mr0grog: It is probably out of scope, but year, I was mainly talking about just these rejection tests. TBH, I was thinking about adding it in the `goto` PR, but felt like it was out of scope there, too. Maybe better as its own separate PR  :P

Mainly it’s just that there should be less rigamarole and better error messaging than this pattern gives us and `mocha-generators` doesn’t seem to patch `should.Throw`, which would be the right generator-based way to do this.

Mr0grog: …but then of course you already went a did it. Sorry I didn’t reply sooner!

rosshinkley: @Mr0grog totally fine! :)

Frankly, I'm not sure that `mocha-generators` should be included considering the "official line" is to use promises instead.  That feels like a topic for conversation elsewhere, though, and may very well be my intentions getting away from me.

That said, `mocha-generators` should (ha) probably patch `throw`.  I'll throw (double ha) that on my list of things to look at.  (Of course, this assumes you haven't done it already.  I haven't set up the ol' @Mr0grog filter on `mocha-generators` yet. ;) )

rosshinkley:
The problem is now fixed via #548, but this PR adds the unit tests for failed selectors.

Issue #375, opened 2015-12-09
waitTimeout does not work


cevek:

sporkmonger:
:+1: I'm also seeing this.

matthewmueller:
can you try again? i just released a new version

cevek:
Cool :+1:

contra:
Also seeing this :rotating_light:

matthewmueller:
@contra ahh yah, you're seeing it hang on the initial page load, right? right now it's only initialized when you call `.wait()`. that should be fixed.

YueShuXie:
Also seeing this.

joychester:
also see this hang forever for own my negative tests(On Mac OS Captian EI + 2.3.2 release), when wait a *non-exist element on page* or wait a *function which does not return true condition* (BTW Page load has already done), 
Please let me know if need my help on reproducing this issue.

rosshinkley:
@contra Depending on the kind of page load failure, #553 _might_ address your issue.

@joychester I wrote two quick test cases:

**nonexistent element**
```js
var Nightmare = require('nightmare'),
  nightmare = Nightmare();

  nightmare.goto('http://www.example.com')
    .wait('does-not-exist')
    .end()
    .then(function(){
      console.log('done');
    })
    .catch(function(err){
      console.dir(err);
    });
```

**function that never returns true**
```js
var Nightmare = require('nightmare'),
  nightmare = Nightmare();

  nightmare.goto('http://www.example.com')
    .wait(function(){return false;})
    .end()
    .then(function(){
      console.log('done');
    })
    .catch(function(err){
      console.dir(err);
    });
```

Both error out with the appropriate error message after 30s.  Can you provide a minimal example that causes nightmare to hang?

joychester:
Hi, @rosshinkley , your case works fine from my side too, but my test pages is not kind of different with yours :)
my page will load the page content async(like SPA application), so if i write wait method like following, it will hang forever:
```javascript
var Nightmare = require('nightmare'),
  nightmare = Nightmare({waitTimeout: 10000, show: true, openDevTools: false});

  nightmare.goto('https://www.stubhub.com/')
    .wait('.event-title123') //this element does not exist..
    .end()
    .then(function(){
      console.log('done');
    })
    .catch(function(err){
      console.dir(err);
    });
```

however, if i hard-wait for the page loaded complete, then the wait() works fine until it timedout(10 seconds in my case)
```javascript
[...]

PR #588, opened 2016-04-22
Upgrade to Electron 0.37.7

Mr0grog:
Contains upstream fixes for #375, #563, #572. For the dev tools issues, I think we should still take #576, but this does obviate it on a technical level.

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,7 +20,7 @@
     "debug": "^2.2.0",
     "deep-defaults": "^1.0.3",
     "defaults": "^1.0.2",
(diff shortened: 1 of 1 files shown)


Mr0grog:
Not sure why that test failed. It seems to run fine locally. Hopefully something didn’t break in this Electron release :\

rosshinkley:
Hm.  I've seen similar screwball failures on Circle, even with the `xvfb` patch.  I don't know if that's still the cause, but it may be worth another look.

PR #591, opened 2016-04-23
Bring back support for `javascript:` URLs

### LINKED ISSUE #590: Execute Javascript URL in goto
I have a page that has a javascript function that gets executed to navigate.  In version 2.3.1 of nightmare I could:

.goto('javascript:__doPostBack(\'ctl100$ctl00$MainMenu\',\'DAT\')')

and it navigated successfully.  When I upgraded nightmare to v2.3.3 it stopped. Here is the return error

{ message: 'navigation error',
      code: -1000,
      details: 'unhandled protocol',
      url: 'javascript:__doPostBack(\'ctl00$ctl00$MainMenu\',\'DAT\')' }

Mr0grog:
Fixes #590.

We used to support `javascript:` URLs, but lost that feature in #553. TBH, I’m slightly skeptical on bringing it back, *but* #553 got released in an extra-minor update, so it seems worth fixing since it actually broke some people.

Bonus: this supports page loading data for URLs like `javascript:window.location='http://google.com'`. It could be ever so slightly faster if we didn’t bother to support that. Seemed nice to have, though. ¯\\\_(ツ)\_/¯

Also worth noting: we don’t handle errors especially well this way. It’s not any worse than before, but I wonder if calling `goto('javascript:notAnActualFunction()')` should return error instead of success, which I think we could do by passing JS URLs through the evaluate code and then, if the result is not `undefined`, setting the URL. That’s complicated, though. (You *can* currently capture errors with the `page-error` event, but it’s hard to connect that with the URL.)

### DIFF
--- a/lib/runner.js
+++ b/lib/runner.js
@@ -15,7 +15,7 @@ var fs = require('fs');
 var urlFormat = require('url');
 var FrameManager = require('./frame-manager');
 
-const KNOWN_PROTOCOLS = ['http', 'https', 'file', 'about'];
+const KNOWN_PROTOCOLS = ['http', 'https', 'file', 'about', 'javascript'];
 
 /**
  * Handle uncaught exceptions in the main electron process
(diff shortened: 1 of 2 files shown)


rosshinkley:
This is a case I would not have thought of.

> Also worth noting: we don’t handle errors especially well this way. It’s not any worse than before, but I wonder if calling `goto('javascript:notAnActualFunction()')`  should return error instead of success...

Oh man.  _My eyes._  I hadn't considered this case, either.  I'd be curious to see a proposed implementation for supporting rejection of bad `javascript:` links.

rosshinkley:
(At any rate, +1.  This will be included in the next release.)

Mr0grog:
> I'd be curious to see a proposed implementation for supporting rejection of bad javascript: links.

The more I think about it, the less I want to do it. It’s pretty edge-casey.

Issue #599, opened 2016-04-25
Make version info available via API


Mr0grog:
To help people with debugging and other issues, we should make the versions of:

- Nightmare itself
- Electron
- Chromium

available at runtime via the API. See also #597

Mr0grog:
A quick note on this: getting the versions for Electron/Chromium requires running Electron, so this either needs to be an async call or it will be a potentially slow, blocking call (the first time; we can cache it after that). Any thoughts on which approach is better, @rosshinkley?

rosshinkley:
Could the information be captured during the `app.on('ready')` handler and piped up as part of the response there, then attached to the Nightmare instance inside of the instance construction sugar?  That would negate the need for extra calls, and that uses a trip that's already being made, so we get version info for (almost) free.  Thoughts?

rosshinkley:
Although... that would mean the version info is not immediately available like you might expect.  It'd only be available after Nightmare has entered a ready state, like it would prior to executing a chain.  I was trying to get away with _not_ having a `.version()` (or similar) method, but now I'm not sure that's particularly feasible.

Mr0grog:
Well, maybe the Electron/Chrome version stuff isn’t necessarily needed in API? We *could* just log it whenever starting Electron if people only need it informationally. Not quite as much utility there, but maybe enough.

rosshinkley:
I think I still prefer exposing _some_ way of getting the versions programmatically, and I can't think of a good method for doing that without an explicit method that doesn't require side-effect-y stuff.

That said, adding a debug call for the versions on `ready` is probably worth doing and will cover most (probably an overwhelming majority) of cases.

rosshinkley:
"Fixed" in 2.4.0, closing.

PR #600, opened 2016-04-26
adds `.engineVersions()` to the API

rosshinkley:
Quick attempt to address #599.


Mr0grog:
I would add a `Nightmare.VERSION` for the version of Nightmare. And maybe rename `.versions()` to `.browserVersions()` or `engineVersions()`?

rosshinkley:
Reopened with suggestions from #604.

PR #604, opened 2016-04-27
get version information synchronously

### LINKED ISSUE #599: Make version info available via API

rosshinkley:
Redo of #600 to fix #599.  This method allows for synchronous version retrieval.

The only thing I'm not _real_ crazy about is the debugging of the version information.  That seems to defeat the purpose of using a lazy getter that props up the cached version values.  Thoughts?

### DIFF
--- a/lib/nightmare.js
+++ b/lib/nightmare.js
@@ -44,6 +44,11 @@ var runner = path.join(__dirname, 'runner.js');
 var template = require('./javascript');
 
 /**
+ * Version
+ */
+Nightmare.version = require(path.resolve(__dirname, '..', 'package.json')).version;
+
+/**
  * Initialize `Nightmare`
  *
  * @param {Object} options
@@ -92,20 +97,11 @@ function Nightmare(options) {
   });
 
   Object.defineProperty(this, 'engineVersions', {
-    get: function() {
-      delete this.engineVersions; 
-      return this.engineVersions = JSON.parse(
-        proc.execSync(`${electron_path} ${runner} --engineVersions`, {
+    value:  JSON.parse(proc.execSync(`${electron_path} ${runner} --engineVersions`, {
           encoding:'utf-8',
           stdio: [null, null, null]
-        }));
-    },
+        }))
   });
-  this.version = require(path.resolve(__dirname, '..', 'package.json')).version;
-
-  debug(`Chromium version: ${this.engineVersions.chrome}`)
-  debug(`Electron version: ${this.engineVersions.electron}`)
-  debug(`Nightmare version: ${this.version}`)
 
   attachToProcess(this);
 
--- a/lib/nightmare.js
+++ b/lib/nightmare.js
@@ -17,7 +17,7 @@ var default_electron_path = require('electron-prebuilt');
 var source = require('function-source');
 var proc = require('child_process');
 var actions = require('./actions');
-var join = require('path').join;
+var path = require('path');
 var sliced = require('sliced');
 var child = require('./ipc');
 var once = require('once');
@@ -35,7 +35,7 @@ module.exports = Nightmare;
  * runner script
  */
 
-var runner = join(__dirname, 'runner.js');
+var runner = path.join(__dirname, 'runner.js');
 
 /**
  * Template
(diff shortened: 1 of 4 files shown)


lib/nightmare.js
```diff
@@ -91,6 +91,22 @@ function Nightmare(options) {
     debug('electron child process exited with code ' + code + ': ' + help[code]);
   });
 
+  Object.defineProperty(this, 'engineVersions', {
```
Mr0grog: Wouldn’t this make more sense of the Nightmare constructor rather than on an instance? (Same for `.version`)

lib/nightmare.js
```diff
@@ -91,6 +91,22 @@ function Nightmare(options) {
...
+    get: function() {
+      delete this.engineVersions; 
+      return this.engineVersions = JSON.parse(
```
Mr0grog: If I’m getting this right, you’re intending to wipe out the getter function and replace it with a static value. Is that right? If so, this doesn’t work—you can only `delete` if `configurable` is `true` in the property descriptor (it defaults to false) and you can only re-set the value with the `=` operator if `writable` is true (also defaults to false). The `=` operator still returns the value you tried to set even if it didn’t work, so the failure isn’t always obvious (one of those really not-so-great design decisions in JS).

lib/nightmare.js
rosshinkley: Possibly, but instances can override electron versions at construction time.  Maybe I'm not understanding what you're getting at?

Mr0grog: > instances can override electron versions at construction time

Ohhh, good point. Did not think about that at all. Still relevant for `.version` maybe?

lib/nightmare.js
rosshinkley: > also defaults to false

Well.  Today I learned `defineProperty` does not work like it does in my mind.  Filing that away.

> The = operator still returns the value you tried to set even if it didn’t work, so the failure isn’t always obvious (one of those not-so-great design decisions in JS).

But the assignment ought to work with `configurable: true`, yes?

Mr0grog:
> The only thing I'm not real crazy about is the debugging of the version information. That seems to defeat the purpose of using a lazy getter that props up the cached version values.

Would agree. Maybe not really necessary at all. If you still want the logging, the electron process could just `parent.emit('log', [versions])` when it starts.

Other note: if `engineVersions` is only available on a Nightmare *instance,* maybe no great need for `execSync`. I know I kind of lobbied for this over on Gitter, but that’s because I was thinking of it being on the Nightmare constructor/the library exports, where it was less tied to an instance.

lib/nightmare.js
Mr0grog: > But the assignment ought to work with configurable: true, yes?

Nope. `configurable` is just about things that change the existence of the property or its metadata (i.e. the other property descriptor fields). For the assignment operator to work, `writable` has to be true.

Mr0grog: (Also you can’t have `writable: true` if you have a getter or setter. So you need to actually call `defineProperty` again or just keep the getter and not call `execSync`)

rosshinkley: I can buy configurable having to be set, that makes sense to me.  I don't know that I buy having `writable: true` is needed.  Maybe I am not understanding this right (_entirely possible_), but `writable` would allow the getter property to be assigned to, and that shouldn't matter if `configurable` is true and the getter is deleted first (which here, it is, and allowable).

rosshinkley:
> Maybe not really necessary at all. 

I was heading in this direction, at least for this incarnation.

> Other note: if engineVersions is only available on a Nightmare instance, maybe no great need for execSync.

I'm kind of 50/50 on this.  I kind of like having the version property be a regular ol' synchronous property.  However, the fact that versions can change by instance makes me think maybe it should be a chainable like it was in #600 (although it would still need some naming and placement touchups).

lib/nightmare.js
Mr0grog: Right, `delete` will remove the property, so you can then write to it because there’s no getter there. You’d only need `writable` to be `true` if the `delete` call weren’t there. But the `delete` call doesn’t work if `configurable` isn’t explicitly set to `true`.

Mr0grog:
```
var thingamajigger = {};
Object.defineProperty(thingamajigger, 'thingy', {
  get: function() { return 'from getter'; }
});
console.log(thingamajigger.thingy);
delete thingamajigger.thingy;
console.log(thingamajigger.thingy);
// logs “from getter"
```

rosshinkley:
... also to add another possible wrench to the mix, and maybe another point for the original incarnation, this will cause a process to never quit:

```js
var nightmare = Nightmare();
console.dir(nightmare.engineVersions);
```
... as the child process is started at construction time.

lib/nightmare.js
rosshinkley: Yeah, that's what I was trying to get at.  It works here by virtue of torching the property, which won't work without `configurable`.  I think (hope?) we agree. :)

Mr0grog: Yep, you’re just missing `configurable: true`.

rosshinkley:
For completeness, I've gone ahead and finished this PR with suggested changes.  However, in light of Electron versions being instance-dependent and the mere instantiation of Nightmare creating a possible "hang", I think it's best to re-open #600.

PR #609, opened 2016-04-28
Proposing nightmare-promise

Oceanswave:
Hello!

This pull is the result of doing a pretty extensive rewrite of nightmare. The readme.md describes some of the new functionality introduced, however, beyond using promises and the auto-chain (fluent) functionality as well as rebasing on 'Nightmare.prototype', there are other dev-centric goals that I wanted to get into the codebase. These are using ecmascript 5 features when it made sense, reducing boilerplate code, modularizing when possible, and reducing dependencies to well known and maintained libraries.

I've been forward porting pulls when they were committed to nightmare. While the commits aren't 1:1 the functionality should be close nee identical to what is in the base fork.

If you would, please check it out and see if these changes align with Nightmare's goals.

Thanks!


LinusU:
I personally wouldn't mind dropping the `.chain()` altogether. When using with `async`/`await` (via babel) or with [co](https://github.com/tj/co) it's seldom needed.

e.g.

```javascript
async function browserGoogle () {
  const browser = new Nightmare()

  await browser.goto('http://yahoo.com')
  await browser.type('input[title="Search"]', 'github nightmare')
  await browser.click('.searchsubmit')
}
```

or

```javascript
co(function * () {
  const browser = new Nightmare()

  yield browser.goto('http://yahoo.com')
  yield browser.type('input[title="Search"]', 'github nightmare')
  yield browser.click('.searchsubmit')
})
```

When writing code that has a lot of evaluates that are used to choose branch it's especially useful.

Oceanswave:
Yeah, I tend to agree - aside from really simple code you're going to have a number of branches which reduces the length of the chain -- plus it leaves the door open for harder to troubleshoot exceptions that may occur somewhere in the chain.

That said, however, it does provide for an easy migration path from the current Nightmare and slightly more readable code when using the fluent, chained, style.

The pattern also has prior precedence in libraries such as lodash that include a chain method for similar readability benefit.

since it's one method, it's easy enough to deprecate in the future...

inca:
Hey guys, can we please have a follow-up on this one? (e.g. when/if could we expect this PR to be merged, can `.chain()` be considered deprecated, etc.) I'd really like to start using fresh new Nightmare with pure async/await env.

matthewmueller:
i'm going to leave it up to the others to decide to merge or not, since i haven't been actively working on nightmare lately, but some quick thoughts:

- the chaining is primarily there to help it so you don't need to compile JS with babel or use co. this is sort of become less and less relevant, i think it's probably fine to just await or yield at each step. you'll probably get better performance when you do want to parallelize this stuff.

- can't you async/await now? nightmare implements thenables, which are basically lazy promises and should give you all the advantages of promises.

- the prototype extension is really clean, but do you get a lot of advantages just using object with functions. if i have some time, i'm planning on having nightmare move more towards this: https://github.com/matthewmueller/plumbing/blob/master/test.js, where you can easily add tracers, logging, queueing (any other decorator) in user-land without cluttering up the fundamental actions of nightmare.

inca:
Thank you for detailed response! I didn't realize that each manipulation returns thenable, thus can basically be used individually. That's exactly what I was looking for.

casesandberg:
Thanks for all of the hard work on this. It's been a while since this was authored and a lot of things have changed. Unfortunately, this doesn't make sense at this time, so I am going to close this.

Issue #607, opened 2016-04-27
Readme Yahoo example needs a `.catch()`


rosshinkley:

shesmu:
the missing catch wouldn't happen to cause a 'Cannot read property 'dock'' would it?

Mr0grog:
@shesmu no, it doesn’t. Are you referring to #602, or are you seeing that error independently (and do you have a reproducible test case)? I can see, where that error is happening, but can’t see how the conditions that cause it would ever be true. I’d love to know if you have a way to make it reliably happen.

rosshinkley:
Fixed in 2.4.0, closing.

PR #613, opened 2016-04-29
Add catch clause and improve selectors in Yahoo example

Mr0grog:
Hopefully this makes the README example a little more robust agains the regular, surface-level changes in Yahoo’s homepage.


Mr0grog:
Also: this fixes #607.

PR #595, opened 2016-04-25
Added hook to append chromium arguments

yotamlaufer:
https://github.com/electron/electron/blob/master/docs/api/app.md#appcommandlineappendargumentvalue

List of arguments:
http://peter.sh/experiments/chromium-command-line-switches/

PR #623, opened 2016-05-06
Add better tests for switches.

### LINKED ISSUE #622: `switches` test improvement
Arising from a conversation over #595, there should be a better test beyond mere construction for `switches`.  One possibility that @Mr0grog suggested is a test-only plugin that hooks into the Electron machinery to ensure the switches are set.

Mr0grog:
Fixes #622.

#595 should be almost exactly the same as ‘should be constructable with switches’ since it is the same, just only supporting switches with no values (or switch + value together in one string).


rosshinkley:
This looks great, adding to next release. :+1:

PR #640, opened 2016-05-17
Add timeout to `goto` action

Mr0grog:
This helps to alleviate the problems noted in #612 (and hinted at in other issues as well), but does not solve the core problem (deadlocking on a disk cache shared between Electron processes).

This turned out to require a much bigger change than one might expect—Nightmare checks whether the window is ready to receive actions (i.e. whether it is loading) between every action, but since we can now be in a state where we are ok to continue but loading is still going on, we have to track for ourselves whether is OK to perform the next action instead of simply asking Electron whether resources are loading. This is accomplished by way of a method named `isReady()` that returns a boolean and `setIsReady(newValue)` that takes a boolean and fires an event on the `BrowserWindow` instance called `did-change-is-ready`.

The actual timeout value can be customized at construction time (via an option called `gotoTimeout`, analagous to `waitTimeout`) and at method time with a third argument `goto(url, headers, timeout)`.

@rosshinkley I’m not 100% providing *both* those methods for overriding is best—maybe it should just follow the example of `wait` and only support the constructor option?

I also wonder if, based on this, certain actions should be exempted from the `continue` check (e.g. `goto`, `viewport`, `wait(number)`), since it’s not really relevant for everything. I think that’s definitely out of scope for this PR, though.


rosshinkley:
> I’m not 100% providing both those methods for overriding is best—maybe it should just follow the example of wait and only support the constructor option?

Hm, I'm inclined to say that it should be a constructor-only option.  I can't think of a case offhand where having differing timeouts by URL would be helpful.  Is there a use case buried here that I'm not thinking of?

> I also wonder if, based on this, certain actions should be exempted from the continue check (e.g. goto, viewport, wait(number)), since it’s not really relevant for everything. I think that’s definitely out of scope for this PR, though.

Fair point.  I suspect the idea was to have all of the actions use the same events for simplicity, but I can certainly see where a go-ahead timeout would be useful for those handful of actions.  (Adding it to `.wait(number)` might be breaking for users in the wild, though.)

Mr0grog:
> I can't think of a case offhand where having differing timeouts by URL would be helpful. Is there a use case buried here that I'm not thinking of?

I *think* this came from my personal angst that it’s not an option with `wait(selector|function)`—where I do think there are clear use cases for different timeouts on different calls (bits of this PR have been hanging around in my git stash for a while). I think, upon reflection, that it doesn’t have any especially great use-cases. Maybe when loading a local page you might expect it to be fast and want to time out more quickly than when loading something from the web? I’m not sure that’s a very strong case, though.

Mr0grog:
Updated to resolve merge conflicts.

rosshinkley:
>I think this came from my personal angst that it’s not an option with wait(selector|function)—where I do think there are clear use cases for different timeouts on different calls (bits of this PR have been hanging around in my git stash for a while).

Ah, I understand.  Wholly agree it makes sense for `.wait()`, but for `.goto()` ... I'm leaning toward "no".

> Maybe when loading a local page you might expect it to be fast and want to time out more quickly than when loading something from the web?

Mmm, _maybe_ - this feels a bit out in the weeds, though.  I say let's not complicate the method signature until it's demonstrably useful.  Thoughts?

Mr0grog:
> Wholly agree it makes sense for .wait(), but for .goto() ... I'm leaning toward "no".

Yeah, already changed it :)

rosshinkley:
@Mr0grog I saw. :)  Doing my best to be thorough.

This looks good.  +1 to you sir.

PR #641, opened 2016-05-17
Add catch function to nightmare instance

LinusU:
It felt very awkward to be able to use `.then`, but not `.catch` directly, leading me to write code like this:

```js
browser.evaluate(checkErrors).then(x => x).catch(err => {
  // ...
})
```

This pr will allow me to skip the `.then(x => x)` part and just use `catch` directly, making it work even more as a normal promise.

```js
browser.evaluate(checkErrors).catch(err => {
  // ...
})
```

### DIFF
--- a/lib/nightmare.js
+++ b/lib/nightmare.js
@@ -377,6 +377,14 @@ Nightmare.prototype.then = function(fulfill, reject) {
 };
 
 /**
+ * catch
+ */
+
+Nightmare.prototype.catch = function(reject) {
+  return this.then(undefined, reject);
+};
+
+/**
  * use
  */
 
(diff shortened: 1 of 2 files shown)


albinekb:
Ping @Mr0grog

Mr0grog:
Cool, I think this would be great to have! But why isn’t the implementation just:

```js
return this.then(undefined, reject);
```

It’s much simpler and more robust. After all, that’s how it’s specified: http://www.ecma-international.org/ecma-262/6.0/index.html#sec-promise.prototype.catch
and even how it’s implemented in V8: https://chromium.googlesource.com/external/v8/+/master/src/promise.js#256

Also, it needs a test ;)

(Side note: I’m not a committer here, just a helpful handyman, so I can’t approve your PR.)

LinusU:
@Mr0grog Thanks for the comments!

rosshinkley:
+1 on this.  Including.

Issue #502, opened 2016-02-25
npm scripts instead of Makefile


fritx:
https://github.com/segmentio/nightmare/blob/master/Makefile
so that we can just run the test on windows seamlessly
makefile seems to be unfriendly to windows

``` js
"scripts": {
  "test": "mocha --harmony"
}
```

rosshinkley:
The makefile buys automatic installation of dependencies if the package has been touched since the last make.  With the inclusion of #561, the makefile will also allow headless systems to run the unit test suite (specifically, CircleCI) without internal problems to Electron/Chromium.  Mimicking these two behaviors with simple NPM scripts would be difficult to do in a cross-platform friendly way (although not impossible).

I'm in favor of modifying the build process to make it friendlier, and would like to get some feedback on what we should do.  Off the cuff, I'd suggest something like Gulp, but that may be a bit heavy-handed for what we ultimately want to accomplish.

Has anyone had trouble building this on Windows or any other OS where the makefile won't work?  If so, have you been able to work around it?  What does your make process look like?  Do you have suggestions for build tools?

Mr0grog:
This is a little tangential, but #571 made me wonder if the scripts added in #565 should really be part of Nightmare itself instead of the tests. That is, using Nightmare as a testing tool on a CI server somewhere seems like a major use-case. As it stands now, every user of Nightmare will have to know that they need to replicate Nightmare’s testing setup. If Nightmare just did all of that on its own before launching Electron, that would not only solve the deadlock for Nightmare’s own tests but also for other tests that leverage Nightmare.

Doing so would also simplify the test script to the point where we might not need to worry about adding Gulp or some other solution to get things working on Windows.

(Note I have no idea whether this is the problem being encountered in #571; that issue just made me think of this.)

rosshinkley:
> If Nightmare just did all of that on its own before launching Electron, that would not only solve the deadlock for Nightmare’s own tests but also for other tests that leverage Nightmare.

I'm going to purposefully totally ignore the NPM automagic make gives and focus on the framebuffer issues for the moment.  I'm tentative on if we should automagically include the "run with `xvfb-runner`" magic.  My main dev box is running both X and Xvfb - what if I want to see the Electron instance?  What should Electron run under?

[...]

PR #644, opened 2016-05-17
Fix deprecated API usage in preload test (needed to support Electron 1.0+)

Mr0grog:
We fixed this in the actual preload script, but missed the test! This doesn’t upgrade Electron, but it is a prerequisite for upgrading. (The API is gone entirely in Electron 1.0)


rosshinkley:
I had spotted this when working on stuff for #502, but hadn't gotten around to PRing it.  Thanks!

Issue #593, opened 2016-04-25
v3 features


rosshinkley:
I'm not sure where to put this, and I'm open to this being moved to a Wiki if that makes more sense.  In the interest of having all of the big (and possibly breaking) features/nice-to-haves for the next big release in one spot:

* Native `iframe` support and management - #203, #496
* Multiple window management support (including popups) - #352, #510 
* IPC safety - #493, #579 
* Queue safety - #493, #239, #522 (and many, _many_ more)

There may be a need for:

* An upgrade guide for users
* Splitting up the documentation (akin to how Electron has it)
* A semi-formal statement of ongoing support for 2.x for some period once 3.x is finished

What have I missed in features, needs, or linked issues?

Mr0grog:
#596 makes me wonder if changing `viewport()` to actually set the *viewport* (as opposed to window) size would be a good breaking change for v3.

Mr0grog:
Maybe have `goto` return as soon as a page is loading so you can execute scripts ASAP during load?

Have actions with selectors wait for the selector before doing the action (#388). Capybara (Ruby), for example, does this. It could possibly be configurable (though I think you’d want opt-out rather than opt-in).

rosshinkley:
IPC safety will likely make it into the next release of Nightmare.

rosshinkley:
> #596 makes me wonder if changing viewport() to actually set the viewport (as opposed to window) size would be a good breaking change for v3.

Hm, how are you planning on doing that?  Using `window.outerWidth` and `outerHeight` and then adding the difference?  (I don't remember being able to set the viewport directly, but I'm probably wrong.)

> Maybe have goto return as soon as a page is loading so you can execute scripts ASAP during load?

To what end?  Should `preload` be extended (or pluggable) instead?

> Have actions with selectors wait for the selector before doing the action (#388). Capybara (Ruby), for example, does this. It could possibly be configurable (though I think you’d want opt-out rather than opt-in).

Fully on-board for an implementation of #388.

Mr0grog:
> how are you planning on [changing the viewport size instead of window size]?

I think just calling [`window.setContentSize()`](http://electron.atom.io/docs/v0.37.8/api/browser-window/#winsetcontentsizewidth-height-animate) instead of [`window.setSize()`](http://electron.atom.io/docs/v0.37.8/api/browser-window/#winsetsizewidth-height-animate)

> To what end? Should preload be extended (or pluggable) instead?

[...]

PR #645, opened 2016-05-18
Make `goto` reject w/ a helpful message when `url` is not provided

### LINKED ISSUE #642: .goto(undefined) throws empty object
```js
const Nightmare = require('nightmare')
const nightmare = new Nightmare()

try {
  nightmare.goto(undefined)
} catch (err) {
  console.log(err)
}
```

Actual output:

```text
{}
```

Expected output:

```text
TypeError: expected string, got undefined
```

Mr0grog:
Fixes #642.

Ultimately, I think it would be better to reject with an instance of `TypeError` (with the same message), but that requires properly serializing and deserializing errors over IPC, which is a non-trivial task (would love to hear your thoughts on this, @rosshinkley). I wanted to get the immediate problem solved first.

### DIFF
--- a/lib/runner.js
+++ b/lib/runner.js
@@ -144,6 +144,10 @@ app.on('ready', function() {
    */
 
   parent.respondTo('goto', function(url, headers, done) {
+    if (!url || typeof url !== 'string') {
+      return done('goto: `url` must be a non-empty string');
+    }
+
     var extraHeaders = '';
     for (var key in headers) {
       extraHeaders += key + ': ' + headers[key] + '\n';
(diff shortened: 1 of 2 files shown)


rosshinkley:
Thank you for this, looks good.

As for the errors, it might be better to move this conversation to #646.  I haven't thought terribly deeply about serializing errors over IPC - my first thought is that all actions would have to consistently handle error cases, which is something we've also touched on in #593.  Once we've got consistent error reporting between the parent and child process, the errors over IPC almost falls out of that for free.  Entirely possible I'm missing something, though.

PR #651, opened 2016-05-20
Give better error messages when element isn't found

LinusU:
Gives more helpful errors when using `.click`, `.mousedown` or `.mouseover`, and the selector cannot be found.

e.g. when running `.click('a.not-here')` and the element cannot be found, it will give back the error "Unable to find element by selector: a.not-here" instead of the previous "Cannot read property 'dispatchEvent' of null".


rosshinkley:
Looks good.  Thanks!

PR #653, opened 2016-05-20
adding engine information

### LINKED ISSUE #621: package should specify engine
Outgrowth of #620.

rosshinkley:
fixes #621.


Mr0grog:
Can we also drop the `--harmony` flags as noted all over the README, then? (I *think* we can, but honestly not totally sure.)

Mr0grog:
May also want to add multiple node versions for testing in `circle.yml` as described here: https://discuss.circleci.com/t/testing-multiple-versions-of-node/542

rosshinkley:
> Can we also drop the --harmony flags as noted all over the README, then? (I think we can, but honestly not totally sure.)

I'm almost positive we can.

> May also want to add multiple node versions for testing in circle.yml as described here: https://discuss.circleci.com/t/testing-multiple-versions-of-node/542

Ah, that's a good call.  I'll have to look into that a little deeper, what do you think is reasonable?  Latest 4, latest 5, latest 6?  Something like that?

Mr0grog:
> what do you think is reasonable? Latest 4, latest 5, latest 6? Something like that?

That’s what I would do :)

If you only specify the main version number with NVM, it’ll use the latest available (e.g. `nvm use 4` will use `4.4.4`).

PR #658, opened 2016-05-23
patching javascript action to remove extra listeners

### LINKED ISSUE #350: Event Emitter Memory Leak: using evaluate function
Hi Nightmare Team

The 1st test uses the evaluate function to retrieve the document.title of the pages.
However, after the 10nth iteration, there is the warning of node of a potential memory leak problem with the event emitters (see screenshot).

The 2nd test shows no such things. Evaluate was removed completly.
Is this the correct behaviour of the evaluate function or do we have a bug here?

1. Test
--------
// with evaluate -> Memory Leak after the 10nth loop
function *start() {
var nightmare = new Nightmare({ show: true });
for (var i = 0; i < links.length; i++) {
var title = yield nightmare.goto(links[i]).evaluate(_extractTitle);
console.log(links[i], title);
}
yield nightmare.end();
}

2. Test
--------
// no evaluate -> everything runs fine!
function *start() {
var nightmare = new Nightmare({ show: true });
for (var i = 0; i < links.length; i++) {
var title = yield nightmare.goto(links[i]);
console.log(links[i], title);
}
yield nightmare.end();
}

Node version: 4.2.2
Nightmare version: 2.14.7

issue

Cheers
R.

rosshinkley:
Fixes #350:  Ensures extra event listeners are removed when done/error is called, also allows for >1 log event per `javascript` call.

All credit should go to @s0m3on3 for the original fix.

(Also, I _thought_ this had already been submitted, but I can't find the original.)

### DIFF
--- a/lib/runner.js
+++ b/lib/runner.js
@@ -248,17 +248,23 @@ app.on('ready', function() {
    */
 
   parent.respondTo('javascript', function(src, done) {
-    renderer.once('response', function(event, response) {
+    var response = (event, response) => {
+      renderer.removeListener('error', error);
+      renderer.removeListener('log', log);
       done(null, response);
-    });
+    };
 
-    renderer.once('error', function(event, error) {
+    var error = (event, error) => {
+      renderer.removeListener('log', log);
+      renderer.removeListener('response', response);
       done(error);
-    });
+    };
 
-    renderer.once('log', function(event, args) {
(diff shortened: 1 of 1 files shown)


rosshinkley:
CircleCI tests are failing because I left my personal build pointed at the 14.04 image to test the fix for #621, which is certainly going to break without the updated circle.yml.

PR #664, opened 2016-05-27
add `.once()` and `.removeListener()`

### LINKED ISSUE #663: .once and .removeEventListener is missing
Only `.on` is implemented, it would be nice if there where a way to stop listening to events, or only listen to the first event.

rosshinkley:
Fixes #663.


LinusU:
Awesome :raised_hands:

PR #673, opened 2016-06-05
Add 'Referer' header support

antixrist:
Hi guys!
How about "Referer" header?

Usage:
```javascript
const url = 'https://www.whatismybrowser.com/detect/what-http-headers-is-my-browser-sending';
var nightmare = new Nightmare();
yield nightmare
  .viewport(1920, 1080)
  .goto(url, {
    'Referer': 'http://my-referer.tld/'
  })
  .wait('body')
  .screenshot(path.join(__dirname, 'referer.png'))
  .end()
;
```

Result:
headers

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@ node_modules
 .DS_Store
 test/tmp
 test/DevTools Extensions
+.idea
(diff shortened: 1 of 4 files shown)


rosshinkley:
Could you include tests, please?  I think that a test verifying the header is set in Express and perhaps another verifying that Electron is returning the correct referrer from `.goto()` would be sufficient.

Somewhat offtopic, I'm not sure how I feel diverging (or diverging more?) from [`webContents.loadURL()`](https://github.com/electron/electron/blob/master/docs/api/web-contents.md#webcontentsloadurlurl-options), but I think I can live with adding `Referer` to the headers in this way.

Also somewhat offtopic, it _might_ be wise to accept both the 3 R (`Referer`) and 4 R (`Referrer`) header names.  Thoughts?

Issue #469, opened 2016-02-10
Control typing speed on .type()


jviotti:
When using `.type()` and showing the Electron window, I can see how the text I put in gets typed. Is there a way to control the typing speed?

matthewmueller:
not right now :-) do you want to speed it up or slow it down?

we have a delay in place as a hack to ensure the characters are entered in the correct order.

jviotti:
@matthewmueller I wanted to speed it up. How much do you think we could increase the speed without compromising the ordering?

kylewelsby:
I've tested my tests with `0` timeout on the `.type()` function and can't report any keystrokes being emitted in the wrong order. 

Using the `setTimeout(fn, 0);` hack to defer the event to the next event loop of JavaScript should be just fine. 

I was tempted to make this timeout a configurable variable to aid debugging without making test suites run longer than necessary, but I did not want to introduce a breaking change.

PR #677, opened 2016-06-08
Speed up `.type()` event

kylewelsby:
closes https://github.com/segmentio/nightmare/issues/469

I've tested my tests with `0` timeout on the `.type()` function and can't report any keystrokes being emitted in the wrong order. 

Using the `setTimeout(fn, 0);` hack to defer the event to the next event loop of JavaScript should be just fine. 

I was tempted to make this timeout a configurable variable to aid debugging without making test suites run longer than necessary, but I did not want to introduce a breaking change.

### DIFF
--- a/lib/runner.js
+++ b/lib/runner.js
@@ -374,13 +374,8 @@ app.on('ready', function() {
         keyCode: ch
       });
 
-      // HACK to prevent async keyboard events from
-      // being played out of order. The timeout is
-      // somewhat arbitrary. I want to achieve a
-      // nice balance between speed and correctness
-      // if you find that this value it too low,
-      // please open an issue.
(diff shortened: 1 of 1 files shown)


coyotte508:
Well this introduces problem when the page doesn't expect things to be typed instantly (like when filling card expiry in paypal's interface)

kylewelsby:
It would be ideal to add the timeout duration as a attribute.

Kind Regards

*Kyle Welsby - MeKyle Ltd*

mobile: +44 (0) 7575 790 552
email: kyle@mekyle.com
twitter: @halfcube <http://twitter.com/halfcube>
site: mekyle.com
skype: halfcubeuk

Issue #665, opened 2016-05-28
could not make the proxy work


mikkokotila:
Is there a code example where the proxy, and proxyAuth options are working. I spent a few hours trying to get it work, but could not. I tried maybe 100 different variations of the syntax as the documentation is not really clear about this, and about half did not return any errors, the remaining code ran fine, but the IP did not change to the one set in the proxy settings. After every attempt I checked the log files on the server, to confirm if requests came in and if the IP was changed. 

Here is where I was left (I've replaced my test domain with google.com below):

    var Nightmare = require('nightmare');

    var testbot = new Nightmare({
   	    switches: {
	    proxy: 'http://1.proxy.com:80', 
	    proxyType: 'http', 
	    proxyAuth : 'user:password'
    }})
    testbot
      .goto('http://google.com')
      .wait()
      .run(function(err, nightmare) {
       if (err) return console.log(err);
      console.log('Done!');
      process.exit()
      });

Mr0grog:
I don’t think any of those are actually valid switches (you can see the whole list at http://peter.sh/experiments/chromium-command-line-switches). I think the switch you want is `proxy-server`:

```js
var Nightmare = require('nightmare');

var testbot = new Nightmare({
  switches: {
    'proxy-server': 'http://1.proxy.com:80'
  }
});
```

I don’t know if you can set the auth info in the URL (e.g. `http://user:pass@1.proxy.com:80`) or if you would need to use the `authentication()` action, e.g:

```js
Nightmare({
  switches: { 'proxy-server': 'http://1.proxy.com:80' }
})
.authentication('user', 'password')
.goto('http://google.com')
.doSomething()
.run(function (error, result) { ... });
```

Alternatively, you could write a plugin that uses the `setProxy` method on Electron’s `session` object, as documented here: http://electron.atom.io/docs/api/session/#sessetproxyconfig-callback

mikkokotila:
Thank you. 

I had tried that as well. I tried again and neither of the examples you change anything i.e. I could change the switch (proxy-server) to lkjlkj-weioruweoriu and the code still runs the rest of the program. It seems to me that for some reason the switches part is totally ignored at the moment, because it can be changed to anything without any effect on the rest of code. Thank you for the plugin tip, though it would be great if this could work as in the examples you had provided given how amazing Nightmare otherwise is!

rosshinkley:
[...]

PR #687, opened 2016-06-16
Update proxy info

agrberg:
Add useful information found in issue #665 for devs looking for proxy info.


agrberg:
I'm not married to the changes in the first two commits so if you don't like them, I'm happy to drop them.

rosshinkley:
Keeping the documentation consistent: could you move the `.evaluate()` and `.then()` methods to be inline?

rosshinkley:
By the by, the first two commits: +1.  Those are logical changes.

agrberg:
Sure thing. I intended that to reduce the duplication but it's minimal and might make things easier on the reader. One sec and I'll have an update.

agrberg:
@rosshinkley updated, thanks for the review

rosshinkley:
Awesome.  Thanks! :)

Issue #704, opened 2016-06-30
Ability to set Timezone


rfink:
I'm actually not 100% sure if this works, but would it be possible to allow environment variables to be set for the child process?  I would like to set TZ to change the timezone that electron is running on.  I can create a pull requests if needed.

rosshinkley:
This should have been fixed with #705.  Closing.

PR #697, opened 2016-06-25
update electron-prebuilt version

rosshinkley:
Fixes #694, also removes an errant `.only()` in the test suite.

PR #705, opened 2016-06-30
Run it.only, and add option to provide env vars to electron process

rfink:
Addresses issue #704, also removes "it.only" from unit tests (assuming that was unintentional).

I admit that I'm not sure what to write for a unit test here.  If you can give me some guidance, that would great.  Otherwise, existing tests pass, and a test run on my machine appears to use the TZ environment variable.  Thanks!

### DIFF
--- a/lib/nightmare.js
+++ b/lib/nightmare.js
@@ -94,7 +94,8 @@ function Nightmare(options) {
   this.queue((done) => {
 
     this.proc = proc.spawn(electron_path, [runner].concat(JSON.stringify(electronArgs)), {
-      stdio: [null, null, null, 'ipc']
+      stdio: [null, null, null, 'ipc'],
+      env: options.env || process.env
     });
 
     this.proc.stdout.pipe(split2()).on('data', (data) => {
--- a/lib/nightmare.js
+++ b/lib/nightmare.js
@@ -22,6 +22,7 @@ var sliced = require('sliced');
 var child = require('./ipc');
 var once = require('once');
 var split2 = require('split2');
+var defaults = require('defaults');
 var noop = function() {};
 var keys = Object.keys;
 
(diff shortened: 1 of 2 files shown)


rosshinkley:
> also removes "it.only" from unit tests (assuming that was unintentional).

Yes, it was unintentional.  Streams got crossed.  (This was also fixed in #697, fwiw.)

> I admit that I'm not sure what to write for a unit test here. If you can give me some guidance, that would great.

I think the easiest thing to do would be to use a custom `.action()` action to pull the environment variable from Electron's `process`.  You should then be able to create an instance with an environment variable, use the custom action to get that variable, and test accordingly.

rfink:
Thanks!  I'll get that going.

rfink:
So what I'm finding, is that the "env" object doesn't get attached to the ChildProcess object in a way that node can access it (i.e. this.child.env, this.proc.env, or this.proc._handle.env).  Unless I'm missing something, it might be hidden under the covers in V8.  Do you have any other suggestions?

rosshinkley:
You should be able to pull `env` off of `process` like you might for any ordinary process, I think.  Off the cuff:

```js

Nightmare.action('envtest',
    parent.respondTo('envtest', function(done){
      done(null, process.env);
    });
    done();
  },
  function(done) {
    this.child.call('envtest', done);
  });

var nightmare = Nightmare();

nightmare.goto('http://example.com')
  .envtest()
  .then((r)=> {
    console.log(r);
  });
```
... should be enough to at least make sure you can get the environment variables back.

rfink:
That's great, thanks!  That makes a lot of sense, I had to learn a little how the process communication was working.  I really appreciate your help and patience.

rosshinkley:
No problem! :)

One minor nit: could you remove the `console.log`?

rfink:
Whoops - my mistake.  Removed.  Thanks!

PR #707, opened 2016-06-30
fix typo in Readme: wit to with

leonstafford:
If the page load fails, the error will be an object **wit** the following properties


rosshinkley:
Thanks for another pair of eyes!

PR #746, opened 2016-08-03
Listen explicitly for provisional load failures in `goto`

### LINKED ISSUE #740: Hangs after 'did-stop-loading' if response is HTTP 304
The process hangs after the 'did-stop-loading' event if the response' HTTP status code is 304 (not modified). I simulated the HTTP 304 response using httpbin (https://httpbin.org/status/304). Here's the Nightmare debug log:

```
  nightmare queuing process start +24s
  nightmare queueing action "useragent" +0ms
  nightmare queueing action "goto" for https://httpbin.org/status/304 +0ms
  nightmare running +1ms
  nightmare:log Navigating: "https://httpbin.org/status/304",
            headers: If-Modified-Since: Sat, 21 Jul 2016 00:00:00 GMT
,
            timeout: 30000 +13ms
  nightmare:log did-start-loading +7ms 
  nightmare:log did-get-response-details +58ms 
  nightmare:log did-fail-provisional-load +2ms 
  nightmare:log did-stop-loading +0ms 
```

There's no timeout or further signal until the process is stopped (Ctrl+C), and then it displays:

    nightmare electron child process exited with code 0: success! +2m

Any clues as to why the process hangs this way.

Mr0grog:
This fixes #740. We used to only listed for `did-fail-load` events to signal failure because they were always emitted in any kind of failure. In Electron v1.2.6, some kinds of provisional failures stopped emitting `did-fail-load`, so we now also need to listen for `did-fail-provisional-load`.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -1,6 +1,7 @@
 dependencies:
   pre:
     - sudo apt-get update; sudo apt-get install libnotify-bin
+    - nvm install 4; nvm install 5; nvm install 6
 test:
   override:
     - nvm use 4 && npm test
--- a/lib/runner.js
+++ b/lib/runner.js
@@ -237,6 +237,7 @@ app.on('ready', function() {
       function cleanup(error, data) {
         clearTimeout(timer);
         win.webContents.removeListener('did-fail-load', handleFailure);
+        win.webContents.removeListener('did-fail-provisional-load', handleFailure);
         win.webContents.removeListener('did-get-response-details', handleDetails);
         win.webContents.removeListener('dom-ready', handleDomReady);
         win.webContents.removeListener('did-finish-load', handleFinish);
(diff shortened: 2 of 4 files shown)


Mr0grog:
See #747 to address test failures in CI—it’s an issue with electron events (again), not an issue with the changes here. Posted it separately because it’ll affect any new PRs :\

Mr0grog:
Rebased on #747 for passing tests.

rosshinkley:
It looks like the `resetChanges` method from 	8015fe0 splices the `cookies` namespace out of existence, causing those tests to fail.

Otherwise, +1.  Thanks, @Mr0grog :)

Mr0grog:
D'oh. Could swear I ran the whole suite locally and didn't see that, but there's no way it couldn't have happened :P

Guess I’m going to have to make that reset algorithm a lot more complicated.

Obviously not germane here, but I wonder if plugins should be made to be either a) uninstallable or b) instance-only.

voidmonk:
@rosshinkley Looks like the fix is passing the tests now. A merge and release will be appreciated.

PR #747, opened 2016-08-03
Stop sending electron event data in forwarded events.

Mr0grog:
Builds for #746 were failing in CI—it turns out this is because JSON serialization is failing from circular references. This is not the first time we’ve seen exceptions on serializing the raw event data, so I’m opting for just removing it altogether here. There’s very little meaningful information that actually comes with them to begin with.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -1,6 +1,7 @@
 dependencies:
   pre:
     - sudo apt-get update; sudo apt-get install libnotify-bin
+    - nvm install 4; nvm install 5; nvm install 6
 test:
   override:
     - nvm use 4 && npm test
--- a/lib/runner.js
+++ b/lib/runner.js
@@ -595,6 +595,8 @@ app.on('ready', function() {
 
 function forward(name) {
   return function (event) {
-    parent.emit.apply(parent, [name, event].concat(sliced(arguments, 1)));
+    // NOTE: the raw Electron event used to be forwarded here, but we now send
+    // an empty event in its place -- the raw event is not JSON serializable.
+    parent.emit.apply(parent, [name, {}].concat(sliced(arguments, 1)));
   };
 }


Mr0grog:
@rosshinkley Whatever you did in https://github.com/segmentio/nightmare/pull/653/commits/03f43fc0b63640d1d05f46434545fcb82c9b2c6a seems to have stopped working for Node 6 :(

Would it be better to `nvm install` in the `pre` scripts?

rosshinkley:
@Mr0grog hm, that's odd.  I'm curious what changed, but in this case, I _imagine_ fixing it first might be simpler.

And yeah, I think a modification to the circle yaml to run `nvm install` for possibly all three versions in the `pre` would work.  Thoughts?

Mr0grog:
> I'm curious what changed, but in this case, I imagine fixing it first might be simpler.

With the events? TBH, I don’t really want to do the dig-through-the-event-and-find-all-the-problems-and-patch-them-then-lobby-electron-for-a-fix-then-see-this-break-later dance again. It seems prone to having serialization issues, and I don’t think that should be a problem Electron cares about—there’s great utility in having some oddball structures and/or exotic objects in the event, just not for Nightmare. Better to leave it alone and only surface the bits that are meaningfully additive here.

> I think a modification to the circle yaml to run nvm install for possibly all three versions in the pre would work. Thoughts?

Yeah, makes sense to me.

rosshinkley:
> With the events? TBH, I don’t really want to do the dig-through-the-event-and-find-all-the-problems-and-patch-them-then-lobby-electron-for-a-fix-then-see-this-break-later dance again.

Ha, I was referring to the Circle setup, but I'm +1 on this too.  I don't _really_ care all that much about including the event data, because as you said, we're not really using the data to begin with.  (I also have an odd sense of deja vu.  Haven't we had this conversation before?)

 Put me down for "strongly agree." :)

> Yeah, makes sense to me.

Cool.  I see you already made the changes.  Thanks!

Mr0grog:
D’oh, sorry to unload when you weren’t even asking about that.

> Haven't we had this conversation before?

Yep, way back in #576 :P

rosshinkley:
> D’oh, sorry to unload when you weren’t even asking about that.

Don't worry about it, I wasn't being terrifically clear.

> Yep, way back in #576 :P

Good gravy.  Well, it's good to know my memory isn't totally failing.

Issue #742, opened 2016-07-30
Feature request: Combine .wait(selector) + .wait(ms) + whichever happens first


nashwaan:
`.wait(selector)` is helpful to wait for an element to appear dynamically in the page upon completion of page loading and the execution of the scripts. So when that element appears, we continue the execution down the nightmare queue. But sometimes we want to `.wait(ms)` in case our `selector` element did not appear rather than throwing exception after `waitTimeout` has passed (default 30 sec).

The ideas is to combine `.wait(selector)` and ` .wait(ms)` and continue whichever happens first.

Either this way:
```Javascript
nightmare
.goto('www.example.com')
.wait('#gallery')
.wait(5000)
...
```
Or this way:
```Javascript
nightmare
.goto('www.example.com')
.wait('#gallery', 5000)
...
```

PR #756, opened 2016-08-12
Fixes segmentio/nightmare#742 - .wait(element, timeout) implementation

noahtkeller:
modified:   lib/actions.js

* map .wait(selector, timeout)->args[1] to this.options.waitTimeout if typeof === 'number'

### DIFF
--- a/lib/actions.js
+++ b/lib/actions.js
@@ -402,7 +402,6 @@ function waitfn() { 
         return done();
       }
       else if (timeout && waitMsPassed > timeout) {
-        debug('.wait() for element timed out after '+timeout+'msec');
         return done();
       }
       else if (self.options.waitTimeout && waitMsPassed > self.options.waitTimeout) {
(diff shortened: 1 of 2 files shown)


rosshinkley:
A couple of explicit tests for this would be nice.  One handling the timeout, one handling the element existing before the timeout should be sufficient.  Thoughts?

noahtkeller:
Sure, I will write them up in a bit.

noahtkeller:
@rosshinkley I have added the test for timeout when no element exists, and it passes. Above you mention "one handling the element existing before the timeout", this test existed already for the .wait(selector) call, so for the .wait(selector, timeout) call I just added a non-existent selector and 150ms to ensure it still passes, it does not pass if I just wait for the ul.

Is there another specific case you want me to test for?

rosshinkley:
@noahtkeller I think this is probably sufficient for now.  Thanks!

Issue #755, opened 2016-08-11
Nightmare persist partition on windows bug


daniel3d:
I was having problems to get Nightmare and Electron app to share a [persist partition](https://github.com/segmentio/nightmare/issues/754)

It turn out is a bug for some reason in windows environment you need to use app.quit() in order to save the persist partition by adding this code

```

/**
 * Make sure we close the app
 */
app.on('window-all-closed', function () {
  // On OS X it is common for applications and their menu bar
  // to stay active until the user quits explicitly with Cmd + Q
  if (process.platform !== 'darwin') {
    app.quit()
  }
})

/**
 * Listen for the app being "ready"
 */
app.on('ready', function() {
  var win, frameManager, options, closed; 

......

```

to runner.js it fixes the problem it may be related with other issues

daniel3d:
Actually only this don't fix the problem it seams the way Nightmare is kilning the process on .end() is not the best way of closing electron process.

ignore the first comment code i posted this one seams to work i cannot find any issue with Nightmare

in runner.js after line 69

```
  /**
   * Exit the electron app correctly
   */
  parent.respondTo('quit', function(done) {
    app.quit(); done();
  });
```

and in the nightmare.js replace line 211 with this
```
    // instead instance.proc.kill(); use quit
    instance.child.call('quit', () =>{
      instance.child.removeAllListeners();
    });
```

rosshinkley:
This should have been fixed with #757, I think.  Closing.

Issue #536, opened 2016-03-15
Sessions are not persisting in Ubuntu


rclai:
The OS is Ubuntu 14.04 on DigitalOcean. I've installed all of these libraries:

```
apt-get update && apt-get install -y xvfb x11-xkb-utils xfonts-100dpi xfonts-75dpi xfonts-scalable xfonts-cyrillic x11-apps clang libdbus-1-dev libgtk2.0-dev libnotify-dev libgnome-keyring-dev libgconf2-dev libasound2-dev libcap-dev libcups2-dev libxtst-dev libxss1 libnss3-dev gcc-multilib g++-multilib
```

I'm leveraging the `xvfb` NPM package to start them up inside my Node script and everything works, except for this:

I'm logging in to this website so that I can re-use the logged in session state for subsequent crawls, but for some reason, the session is not persisting.

I've tested this script on OSX, though without the usage of the `xvfb` NPM package, and it works.

Is there something that `xvfb` does that prevents the session from persisting?

Note: The `enterIframe` is from [this PR](https://github.com/segmentio/nightmare/pull/496).
```js
var Nightmare = require('nightmare');
var Xvfb = require('xvfb');

var xvfb = new Xvfb({ silent: true });
xvfb.startSync();

var nightmare = Nightmare({
  waitTimeout: 5000,
  paths: {
    userData: __dirname + '/sessions',
  },
  webPreferences: {
    images: false,
    partition: 'persist:asdf',
  }
});

var LOG_IN = true; // Change to false when you want to just see if you're logged in

if (!LOG_IN) {
  nightmare
    .goto('https://www.asdf.com/MyAsdf')
    .wait('.header-hello')
    .wait()
    .evaluate(function() {
      return document.body.textContent.match(/Richard/i);
    })
    .end()
    .then(function(result) {
      console.log('result', result);
      xvfb.stopSync();
    }, function(error) {
      console.log('error', error);
      xvfb.stopSync();
    });
} else {
  nightmare
    .goto('https://www.asdf.com/MyAsdf/Login?ReturnUrl=%2fclassic%2fregistereduser%2fmyasdf.aspx%3fsite%3dus%26lang%3den')
    .wait('#frmLogin')
    .enterIframe('#frmLogin')
    .type('#username', 'asdf@asdf.com')
    .type('#password', 'asdf')
    .click('input[type="submit"][value="Log In"]')
    .exitIframe()
    .wait('.header-hello')
    .wait()
    .evaluate(function() {
[...]

Issue #531, opened 2016-03-07
Always fail to set cookie


Velissel:
It seems like setting a cookie always fails. Below is my config object:

`
{
	show: false,
	images: false,
	waitTimeout: 5000,
	paths: {
		userData: path.join('./appData/userData'),
		temp: path.join('./appData/temp'),
		home: path.join(./appData/home')
	}
}
`

What I have done is: 
start a nightmare instance A to url , login to somewhere, extract all cookies as array, close instance A
then start a another instance B, set cookies to extracted ones from instance A

Problem is B.cookies.set(array) always fails...
`
B.cookies.set(array).then(null, function(err) {
  // err is an empty object...
});
`
One of cookie entry has attribute like:
"domain: something.abc.com","hostOnly":true,"path":"/","secure":false,"httpOnly":true,"session":true,"expirationDate":0
It is the one that fails all the time...
Looks like it is because setting cookie to domain 'something.abc.com' when you are not on that page is causing this problem...
Can anyone help please...

rclai:
I am having this issue as well.

rosshinkley:
I don't think you can transfer cookies across domains.  The [current specification](http://tools.ietf.org/html/rfc6265) that most browsers honor says the domain requested from the browser must match the cookie's origin domain or be a subdomain.  I'm paraphrasing, of course, but that's the spirit.  [This article on the current state of cookie domains](http://bayou.io/draft/cookie.domain.html) does a much more comprehensive job explaining how this works than I could do here.

@Velissel's example appears to be trying to put cookies from one domain onto another, which is forbidden to the best of my knowledge.

@rclai _might_ be talking about something different: are you trying to reuse cookies from the same domain across Nightmare instances?

rclai:
@rosshinkley I figured it out, this issue is not relevant to me anymore.

rosshinkley:
@rclai Ah, okay.  Good to know.  Thanks for the response. :)

Velissel:
@rosshinkley, problem is even with .something.abc.com, it does not work. What I wanted to do is as simple as set cookies before I start anything, but still can not find out a way

rosshinkley:
> ... set cookies before I start anything...
Are you trying to set cookies before your first call to `.goto()`?  I don't think `.cookies.set()` will work prior to a page being loaded.

Velissel:
[...]

Issue #398, opened 2015-12-21
How to run nightmare without saved user data: cookies, localstorage, ...?


cevek:

rosshinkley:
Two methods, depending on what you need:

1. Use [`webPrefereces.partition`](https://github.com/atom/electron/blob/master/docs/api/browser-window.md#new-browserwindowoptions).  Example:

```javascript
var nightmare = Nightmare({
   webPreferences:{
       partition: 'your-unique-partition-name'
   }
});
```

2. Use [`options.paths`](https://github.com/segmentio/nightmare#paths).  Example:

```javascript
var nightmare = Nightmare({
   paths:{
       userData: '/path/to/your/data'
   }
});
```

dingziran:
@rosshinkley Could the second method used to share session between different machines?

rosshinkley:
@dingziran _Great_ question.  I'd suspect the answer to that is not a clear "yes" or "no" - it'd probably depend on how strict the site-in-question's security policy is and if it does any sort of request origin validation.

I'll put this one on the list, though.  It's probably worth a peek. :)

Velissel:
@rosshinkley I am new to nightmare, could you please explain how to use partition to share session between multiple instance of nightmare?

rosshinkley:
@Velissel I took a minute to take another look at this.  Provided your instances share partition information, session information is shared across instances.

However, if you open an instance, log in, and open a second instance with the same persistence data before closing the first, the session information doesn't appear to be shared (or the data from the first session isn't persisted or shareable yet).  I'm still digging into that.

Velissel:
@rosshinkley, In my case, persist does not keep session no matter if I start second instance after or before close of first one

albinekb:
Maybe related to https://github.com/electron/electron/issues/5673 ?

PR #757, opened 2016-08-13
Quit electron app gracefully instead of killing the process

daniel3d:
Still not sure what exactly is causing the problem but it seams when you kill() the process is not giving time to electron.app to quit gracefully and that is not allowing the cookies/session to be saved. 

More info can be found here: [Quit gracefully when Ctrl-C is pressed in console #5273](https://github.com/electron/electron/issues/5273)

Fixes bug:
[Nightmare persist partition on windows bug #755](https://github.com/segmentio/nightmare/issues/755) 

Probable fix for:
[Always fail to set cookie #531](https://github.com/segmentio/nightmare/issues/531) 
[sharing a session between different machines? #398](https://github.com/segmentio/nightmare/issues/398)
[Sessions are not persisting in Ubuntu #536](https://github.com/segmentio/nightmare/issues/536)

### DIFF
--- a/lib/nightmare.js
+++ b/lib/nightmare.js
@@ -208,8 +208,9 @@ function endInstance(instance, cb) {
     instance.proc.on('close', (code) => {
       handleExit(code, instance, cb);
     });
-    instance.child.removeAllListeners();
-    instance.proc.kill();
+    instance.child.call('quit', () =>{
+      instance.child.removeAllListeners();
+    });
   } else {
     debug('electron child process not started yet, skipping kill.');
     cb();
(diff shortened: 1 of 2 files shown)


albinekb:
What's the status here?

daniel3d:
Im using it in the moment with this modifications so i can persist session between 2 electron windows. I have noted some of the issues i think will fix but nobody reply from them yet.

rosshinkley:
This change seems reasonable to me, and I wonder if it might effect the outstanding issues with orphaned processes?

Anyway, including.  Thanks!

PR #758, opened 2016-08-14
readme: update mocha example

dunn:
As-is the test will time out.

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -60,25 +60,26 @@ node example.js
 Or, let's run some mocha tests:
 
 ```js
-var Nightmare = require('nightmare');
-var expect = require('chai').expect; // jshint ignore:line
+const Nightmare = require('nightmare');
+const expect = require('chai').expect; // jshint ignore:line
+
+require('mocha-generators').install();
 
 describe('test yahoo search results', function() {
-  it('should find the nightmare github link first', function(done) {
-    var nightmare = Nightmare()
-    nightmare
+  it('should find the nightmare github link first', function* () {
+    const nightmare = Nightmare();
+    yield nightmare
       .goto('http://yahoo.com')
       .type('form[action*="/search"] [name=p]', 'github nightmare')
       .click('form[action*="/search"] [type=submit]')
       .wait('#main')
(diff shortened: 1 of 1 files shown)


dunn:
Actually the test still doesn't work; it "passes" no matter what.  Any idea what the problem there is?

noahtkeller:
Since it's a generator you need to yield at the beginning of the chain. Also, you will no longer need the done() callback, or done variable in the generator declaration.

    yield nightmare
      .goto('http://yahoo.com')

I removed the .end() in the chain as well.

Also, the selector won't return the right results, since the first results are from yahoo image search. The following selector delivers the desired results:

    #main .searchCenterMiddle > li:nth-child(2) a

noahtkeller:
    var expect = require('chai').expect; // jshint ignore:line

    require('mocha-generators').install();

    describe('test yahoo search results', function () {
      it('should find the nightmare github link first', function*() {
        var nightmare = Nightmare();
        yield nightmare
          .goto('http://yahoo.com')
          .type('form[action*="/search"] [name=p]', 'github nightmare')
          .click('form[action*="/search"] [type=submit]')
          .wait('#main')
          .evaluate(function () {
            return document.querySelector('#main .searchCenterMiddle > li:nth-child(2) a').href
          })
          .then(function (link) {
            expect(link).to.equal('https://github.com/segmentio/nightmare');
          })
      });
    });

dunn:
Awesome; PR updated. Thanks!

Mr0grog:
FYI—all the examples in the readme used to use generators and were explicitly changed to *not* use them several months ago. Which is to say, it might be better to make this PR just use promises. (Though it sounds like it still would've needed fixing anyway!)

rosshinkley:
I agree with @Mr0grog.  The examples should probably use plain promises to minimize external dependencies.

dunn:
Gotcha.  I might have time to update this later in the week, but if anyone wants to take care if it first, plesase do.

dunn:
I ended up not using Nightmare for the project that prompted this, so I'm not going to have time to rewrite and test the example code.

Issue #775, opened 2016-09-01
Add XPath support


gjmveloso:
Hi,

I was thinking on how we can add XPath selectors support on nightmare.js and I'd tried the easy path - relying directly on ```$x()``` helper, available on Chrome. 

Not sure why it's not working within Electron/webContents.

Thoughts?

Thanks,
Gustavo

rosshinkley:
According to the [Command Line API Reference](https://developers.google.com/web/tools/chrome-devtools/debug/command-line/command-line-reference):

> This API is only available from within the console itself. You cannot access the Command Line API from scripts on the page.

In other words, `$x()` isn't available from an evaluation context.

You could use `document.evaluate()`, it takes an XPath expression as an argument.  [MDN](https://developer.mozilla.org/en-US/docs/Introduction_to_using_XPath_in_JavaScript) has a good introduction on this topic.

PR #780, opened 2016-09-05
Initial support for XPath

gjmveloso:
Adding `document.evaluate()` before regular `document.querySelector()` calls, transparently handling XPath expressions. 

Could be an initial implementation to target #775 

Thanks,
Gustavo

### DIFF
--- a/lib/actions.js
+++ b/lib/actions.js
@@ -58,7 +58,7 @@ function xpathLookup(xpath) {
         if (e instanceof DOMException) {
             return undefined;
         } else {
-            throw new Error(e.message + " ('" + xpath + "')");
+            throw e;
         }
     }
 }
@@ -76,7 +76,7 @@ function cssLookup(selector) {
         if (e instanceof DOMException) {
             return undefined;
         } else {
-            throw new Error(e.message + " ('" + selector + "')");
+            throw e;
         }
     }
 }
--- a/lib/actions.js
+++ b/lib/actions.js
@@ -11,41 +11,6 @@ var fs = require('fs');
 var keys = Object.keys;
 
 /**
- * Get the version info for Nightmare, Electron and Chromium.
- * @param {Function} done
- */
-exports.engineVersions = function(done) {
-    debug('.engineVersions()');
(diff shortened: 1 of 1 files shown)


gjmveloso:
Hi @rosshinkley,

Can you help me on this? I'm receiving the method ```cssLookup``` is not defined and I'm pretty sure I'm missing something obvious. An outside eye could help me a  lot.

rosshinkley:
A quick look at this - you're trying to call `cssLookup` from within an `evaluate_now`ed function when the function is defined outside of the evaluated context.  That closed-over scope is not carried to the evaluation as it crosses the IPC boundary.  (See [here](https://github.com/rosshinkley/nightmare-examples/blob/master/docs/common-pitfalls/variable-lifting.md) for a slightly longer explanation.)  You'll probably want to tuck that function into the preload.

That said... I'm not _entirely_ sure this is necessary.  Is there a problem that XPath solves that CSS selectors cannot?  More generally, are there users out there that this feature would be of use to?

gjmveloso:
Hi @rosshinkley,

Thanks for your response. XPath is the only available option to select elements by the inner text, for instance.

It's widely used on Selenium space, for example.

Thanks,
Gustavo

jeffijoe:
Please make this happen 😄

lib/actions.js
```diff
@@ -11,12 +11,48 @@ var fs = require('fs');
...
+
+/**
  * Get the version info for Nightmare, Electron and Chromium.
  * @param {Function} done
  */
-exports.engineVersions = function(done){
-  debug('.engineVersions()');
-  done(null, this.engineVersions);
+exports.engineVersions = function(done) {
+    debug('.engineVersions()');
```
justinmchase: You should probably adjust your code editor to be two spaces and undo all of the whitespace formatting changes in this file.

reinpk:
@gjmveloso as @rosshinkley mentioned we've decided to keep css lookup for minimal simplicity. if there's a super clean, tested and ready-to-go PR may reconsider!

PR #781, opened 2016-09-06
.wait(fn, timeout, fnArgs...) implementation

noahtkeller:
modified:   lib/actions.js
- Set soft timeout if second argument passed with fn is typeof number

modified:   test/index.js
- Added test for .wait(fn, timeout) - Test fails if no timeout is passed
- Added test for .wait(fn, timeout, fnArgs...) - Test fails if no timeout is passed

### DIFF
--- a/lib/nightmare.js
+++ b/lib/nightmare.js
@@ -319,6 +319,11 @@ Nightmare.prototype.run = function(fn) {
   this._queue = [];
   var self = this;
 
+  // Did we pass a valid callback?
+  if (typeof fn !== 'function') {
+    fn = function() {};
+  }
+
   // kick us off
   next();
 


noahtkeller:
This only works if the number of arguments passed to the .wait() method is 2 greater than the number of arguments that the wait fn accepts.

The following will timeout after nightmare's waitTimeout is reached:
```
nightmare
    .wait(function(number, arg) {
      return arg;
    }, 1500, false);
```

Where this will soft timeout after 1500msec
```
nightmare
    .wait(function(arg) {
      return arg;
    }, 1500, false);
```

noahtkeller:
The tests are passing locally, and it's showing failed on a part I don't think I interfered with.

noahtkeller:
Stale

Issue #791, opened 2016-09-14
__nightmare is undefined at wellsfargo.com


drew-wallace:
OS: Windows 10
node: 5.10.1
nightmare: 2.7.0
electron: electron-v1.3.6-win32-x64

Navigate to wells fargo and try to do anything.

``` javascript
var Nightmare = require('nightmare');
var nightmare = Nightmare({ show: true })

nightmare
   .goto('https://wellsfargo.com')
   .evaluate(function () {
      return "Made it";
   })
   .end()
   .then(function (result) {
      console.log(result)
   })
   .catch(function (error) {
      console.error('Login failed:', error);
   });
```

The electron window will come up with wellsfargo.com loaded and do nothing.
Open the console and you will see two errors, both saying `__nightmare is not defined`

I tested the yahoo search example and it worked fine. My guess is that it's forgetting the injection that happened.

Further investigation:
If you inspect the `window` object you will notice `__nighmare` no longer exists under `window.__nightmare`, it is now under `window[""].nightmare`

drew-wallace:
I changed both of the ipc declarations in javascript.js to:
```javascript
var ipc = (window.__nightmare ? __nightmare.ipc : window[''].nightmare.ipc);
```
And it works now.

drew-wallace:
Making a PR...

1mike12:
huh, _5300_WellsFargoEmployees also returns undefined

reinpk:
merged PR

PR #795, opened 2016-09-17
Fix for #791 __nightmare is undefined on wellsfargo.com

drew-wallace:
This might be a good fix for any site that has issues with the code injection.

Fix for #791


reinpk:
seems slightly bizarre that this is necessary, but browsers ¯\_(ツ)_/¯ thanks!

PR #796, opened 2016-09-18
Remove `object-assgin` unnecessary object-assign dependency

antpaw:


rosshinkley:
All for removing unnecessary dependencies.  Thanks!

Issue #799, opened 2016-09-21
wait() broken in ES Strict Mode


sram1337:
Calling `nightmare.wait('#someID')` while executing in Strict Mode (ie `node --use_strict`) causes the following error:

```js
/Users/samramirez/Projects/loj/node_modules/nightmare/lib/actions.js:384
  waitfn.apply(this, [self, elementPresent, done]);
                            ^

ReferenceError: elementPresent is not defined
    at Object.waitelem (/Users/samramirez/Projects/loj/node_modules/nightmare/lib/actions.js:384:29)
    at Nightmare.exports.wait (/Users/samramirez/Projects/loj/node_modules/nightmare/lib/actions.js:347:14)
    at Nightmare.next (/Users/samramirez/Projects/loj/node_modules/nightmare/lib/nightmare.js:315:12)
    at /Users/samramirez/Projects/loj/node_modules/nightmare/lib/nightmare.js:323:46
    at EventEmitter.<anonymous> (/Users/samramirez/Projects/loj/node_modules/nightmare/lib/ipc.js:93:18)
    at EventEmitter.g (events.js:273:16)
    at emitNone (events.js:80:13)
    at EventEmitter.emit (events.js:179:7)
    at ChildProcess.<anonymous> (/Users/samramirez/Projects/loj/node_modules/nightmare/lib/ipc.js:49:10)
    at emitTwo (events.js:100:13)
```

I will include code that can reproduce the error below.

The issue is that `wait` calls an `eval` which tries to create a variable in the local scope, and then use that variable.

actions.js:375
```
function waitelem (self, selector, done) {
  eval("var elementPresent = function() {"+
      "  var element = document.querySelector('"+jsesc(selector)+"');"+
      "  return (element ? true : false);" +
      "};");
  waitfn.apply(this, [self, elementPresent, done]);
}
```

When executing in Strict Mode, `eval` can't create local variables [(see section 1.1 here)](http://www.2ality.com/2014/01/eval.html).  So after the `eval`, `elementPresent` doesn't exist.

I submit a [PR](https://github.com/segmentio/nightmare/pull/800) for a fix.  I think its as easy as declaring `elementPresent` before calling `eval`.

Thanks,
Sam

Edit:

breaksOnStrict.js
```
var Nightmare = require('nightmare');
var nightmare = new Nightmare();
nightmare
    .goto('https://github.com/segmentio/nightmare')
    .wait('body')
    .end()
    .then(() => {
        // do stuff
    });
```

[...]

PR #800, opened 2016-09-21
799 - Fixed wait() on Strict Mode

sram1337:
Fixes ReferenceError when called in Strict Mode.

see issue #799


rosshinkley:
At some moment, we should _probably_ prop up linting as a part of the build process (see long-standing #364), but I won't hold up this PR for that.  Including.  Thanks!

PR #807, opened 2016-09-25
Add a document for mouseover(selector)

38elements:
This adds a document for [mouseover(selector)](https://github.com/segmentio/nightmare/blob/2c3284723bc5c7d82a64ac8130af3b84774e4d54/lib/actions.js#L126-L137).


rosshinkley:
Ah, this must have gotten overlooked.  Thanks!

PR #815, opened 2016-09-29
Fix visible to work as expected

ligaz:
Fix `visible` action to check against the current viewport using `getBoundingClientRect`.
You can check [here](http://stackoverflow.com/a/7557433/6409) for more information.

### DIFF
--- a/lib/actions.js
+++ b/lib/actions.js
@@ -56,8 +56,16 @@ exports.visible = function(selector, done) {
   debug('.visible() for ' + selector);
   this.evaluate_now(function(selector) {
     var elem = document.querySelector(selector);
-    if (elem) return (elem.offsetWidth > 0 && elem.offsetHeight > 0);
-    else return false;
+    if (elem) {
+      var rect = elem.getBoundingClientRect();
+      return (
+        rect.top >= 0 &&
+        rect.left >= 0 &&
+        rect.bottom <= (window.innerHeight || document.documentElement.clientHeight) &&
+        rect.right <= (window.innerWidth || document.documentElement.clientWidth)
+      );
+    };
+    return false;
   }, done, selector);
 };
 
--- a/lib/actions.js
+++ b/lib/actions.js
@@ -59,6 +59,8 @@ exports.visible = function(selector, done) {
     if (elem) {
       var rect = elem.getBoundingClientRect();
       return (
+        rect.width > 0 &&
+        rect.height > 0 &&
         rect.top >= 0 &&
         rect.left >= 0 &&
         rect.bottom <= (window.innerHeight || document.documentElement.clientHeight) &&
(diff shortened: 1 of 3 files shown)


ligaz:
@rosshinkley what's the state of this PR? 

Do you plan to merge it or you are afraid it will be breaking change?

rosshinkley:
Last I saw, the tests for `.visible()` were broken.  I haven't had a chance to really dig into it since.

At first blush, this looks like a reasonable change, and also looks like it adds a check for elements that are floated off into oblivion, which is a welcome addition.  I'll take a closer look when I have a chance, but on the face of it, this looks good.

rosshinkley:
On slightly closer inspection: what about elements that are bigger than the viewport?  Doesn't this assume that to be visible, the bottom (and rightmost edge) of the element has to be in view?  I don't think a call to `.visible()` implies the _entire_ element is visible.  In fact, all `.visible()` really implies is that the element is on the page and rendered.  Changing that would be a _substantial_ change and probably a breaking one at that.

ligaz:
The definition of `visible` (at least mine) is that the user can interact with the element in the current view port. This implies that if the element is bigger than the viewport, but some parts of it are present in the viewport it should be `visible`.

`Doesn't this assume that to be visible, the bottom (and rightmost edge) of the element has to be in view? ` - Yes. Thinking twice about it maybe we should check the left top edge instead. Should I update the PR accordingly?

`In fact, all .visible() really implies is that the element is on the page and rendered.` - Isn't this the purpose of `.exists()`?

rosshinkley:
> The definition of visible (at least mine) is that the user can interact with the element in the current viewport. This implies that if the element is bigger than the viewport but some parts of it are present in the viewport it should be visible.

I don't think what you have will work like that.  Specifically:

```js
 rect.bottom <= (window.innerHeight || document.documentElement.clientHeight) &&
 rect.right <= (window.innerWidth || document.documentElement.clientWidth)
```
... will dictate that the bottom and rightmost edge _must_ be in the viewport to be considered visible, and (correct me if I'm wrong) I don't think that's your intent.

>  Thinking twice about it maybe we should check the left top edge instead. Should I update the PR accordingly?

I don't think relying on an edge (or more than one edge) being in the viewport is a good indicator of visibility, as the element you are checking may extend beyond the viewport in any, a combination, or every direction.  For this to work right, I think you'll want to make sure that a part of the viewport overlaps the client rectangle, edge-inclusive.

>  Isn't this the purpose of .exists()?

In short, no.  `.exists()` ensures the element is a part of the DOM, visible or not.  `.visible()` makes sure the element is actually rendered.  I don't think whether or not the element is in the viewport matters, and I would not be surprised if this quirk (for want of a better word) of implementation is relied upon by users of Nightmare.

Pondering out loud: this _might_ be better served as a plugin.

ligaz:
Thanks, I'm closing this PR for now.

PR #819, opened 2016-10-04
asynchronous evaluate

### LINKED ISSUE #552: Wait for promise to settle in document

rosshinkley:
This PR is intended to let you run `.evaluate()`d functions that are asynchronous, either with callbacks or with promises.  For callbacks, if the `.evaluate()`d function's arity is one less than the passed in parameters, it'll assume the last argument to the function is the callback.  If the return value is a `then`able, it'll call `then()` to wait for promise fulfillment.  Otherwise, the call will behave synchronously as it does now.

This change is _potentially_ breaking as it does change behavior based on the arguments to `.evaluate()`.  Is this change worth the risk of possibly breaking implementations that already exist?

Related to #573 and closes #552.

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -134,6 +134,14 @@ var nightmare = Nightmare({
   loadTimeout: 1000 // in ms
 });
 ```
+##### executionTimeout (default: 30s)
+The maxiumum amount of time to wait for an `.evaluate()` statement to complete.
+
+```js
+var nightmare = Nightmare({
+  executionTimeout: 1000 // in ms
+});
+```
 
 ##### paths
 The default system paths that Electron knows about. Here's a list of available paths: https://github.com/atom/electron/blob/master/docs/api/app.md#appgetpathname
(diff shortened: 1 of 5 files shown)


xat:
This feature would be really awesome!

bchr02:
@rosshinkley this pull request is great! Thank you. It will help me with https://github.com/segmentio/nightmare/issues/830

rosshinkley:
Including.  Thanks for the feedback. :)

TomAshley303:
I think this PR can help me. Can someone please show me an example of how to resolve this? https://github.com/segmentio/nightmare/issues/1279

Issue #824, opened 2016-10-11
Nightmare process won't exit if page has confirm dialog on page exit


Fauntleroy:
When trying to `.end()` a nightmare instance the process will hang forever if there is a `confirm` dialog on page exit. The confirm dialog is also not visible.

Code:

``` js
var Nightmare = require('nightmare');
var nightmare = Nightmare({ show: true });

nightmare
  .goto('https://docs.google.com/forms/d/e/1FAIpQLSe2SK5Vzy82yB9SjLI5B3zfrR1QEaxyjyRGvVxWp_K66p31ZA/viewform')
  .type('input[name="entry.1921379240"]', 'NodeSchool Oakland')
  .end()
  .then(function () {});
```

My view of the electron window:

![screen shot 2016-10-10 at 6 29 43 pm](https://cloud.githubusercontent.com/assets/507047/19255990/d89529d4-8f17-11e6-8a70-d1f84d72fd8f.png)

Is there some way to forcefully close the nightmare instance?

Fauntleroy:
I have isolated the issue. It seems that Google Forms is setting the `unload` event and preventing navigation with the special `unload` confirm dialog.

I have a temporary fix right now:
- Hijack `window.addEventListener` and silently kill attempts to bind `unload` events
- Make `window.onunload` and `window.onbeforeunload` unwritable properties

I'll probably submit a pull req after I've cleaned up my solution.

rosshinkley:
For what it's worth, you could instead prop up a custom preload to handle the `confirm` case manually.  By default, the confirm handler returns the default response, which in this case is _probably_ to stay on the page.

Fauntleroy:
@rosshinkley in this case, `window.confirm` isn't the issue, it's the `unload` and `onbeforeunload` handlers. I managed to fix it with a custom `preload.js` script, which I've submitted in a pull request here: https://github.com/segmentio/nightmare/pull/825

rosshinkley:
Fixed with #825.  Closing.

PR #825, opened 2016-10-11
Prevent unload dialogs (...from preventing .end())

Fauntleroy:
Earlier today I ran into this issue: https://github.com/segmentio/nightmare/issues/824

I resolved it by creating a modified `preload.js` that prevents the browser from setting any `unload` or `beforeunload` handlers. This pull request includes my modified `preload.js` along with two tests that ensure that it works.

### DIFF
--- a/lib/preload.js
+++ b/lib/preload.js
@@ -8,6 +8,29 @@ window.addEventListener('error', function(e) {
 });
 
 (function(){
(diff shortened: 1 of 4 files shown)


rosshinkley:
This looks like the same issue that alert, confirm, etc poses.  I'll fix the test conflict and pull it in.

Issue #897, opened 2016-11-20
README clearCache example not working for me


mzuneska:
I was just trying to recreate the [`clearCache` action in the README](https://github.com/segmentio/nightmare#extending-nightmare) and have been unable to get a simple mocha test to pass. Any help is appreciated.

using:
mocha 3.1.2
nightmare 2.8.1
node 4.2.4

Calling `DEBUG=nightmare mocha test.js` results in
```
  test a basic defined action
  nightmare queuing process start +0ms
  nightmare queueing child action addition for "clearCache" +1ms
  nightmare queueing action "clearCache" +1ms
  nightmare queueing action "goto" for http://example.org +0ms
  nightmare running +0ms
    1) should run a simple test with clear cache


  0 passing (30s)
  1 failing

  1) test a basic defined action should run a simple test with clear cache:
     Error: timeout of 30000ms exceeded. Ensure the done() callback is being called in this test.
```

test.js code
```
var Nightmare = require('nightmare');

Nightmare.action('clearCache',
  function(name, options, parent, win, renderer, done) {
    parent.respondTo('clearCache', function(done) {
      win.webContents.session.clearCache(done);
    });
    done();
  },
  function(message, done) {
    this.child.call('clearCache', done);
  });

describe('test a basic defined action', function() {
  this.timeout(30000);
  it('should run a simple test with clear cache', function(done) {
    Nightmare()
      .clearCache()
      .goto("http://example.org")
      .end()
      .then(function() {
        console.log("done");
        done();
      })
  });
});

**Comment by mzuneska** [comment]
The second function signature is incorrect. The code should read as the following.

```
var Nightmare = require('nightmare');

Nightmare.action('clearCache',
  function(name, options, parent, win, renderer, done) {
    parent.respondTo('clearCache', function(done) {
      win.webContents.session.clearCache(done);
    });
    done();
  },
  function(done) {
    this.child.call('clearCache', done);
  });

describe('test a basic defined action', function() {
  this.timeout(30000);
  it('should run a simple test with clear cache', function(done) {
    Nightmare()
      .clearCache()
      .goto("http://example.org")
      .end()
      .then(function() {
        console.log("done");
        done();
      })
  });
});
```

PR #898, opened 2016-11-20
update README to reflect correct signatures

mzuneska:
action parent function only takes one argument

Fixes https://github.com/segmentio/nightmare/issues/897


rosshinkley:
Good eye.  Thanks!

Issue #575, opened 2016-04-15
Should `run()` exist?


Mr0grog:
This is spawned from discussion on #501. @rosshinkley notes:

> `.run()` is really intended for internal use and usage isn't directly supported. Should it be?
1. `run` is not documented, it’s not the “official” API, and nothing internal ever calls it except `end`, which could just as easily call `then` instead. If `run` isn’t intended for external usage, it should be removed (or at least deprecated by logging warnings to the console when called until v3 is released).
2. On the other hand, even though it’s not documented, “official,” or public, I get the impression that it _is_ widely used. It comes up in issues a lot. Does that mean it should _become_ documented and official?

I can see a lot of value in simply canonizing a thing lots of people seem to be using. That said I personally like a smaller API surface, so I think I lean towards deprecating `run`. In the end, though, I’m not sure I have a real horse in the race here. I’d love to hear others’ thoughts.

rosshinkley:
> run is not documented, it’s not the “official” API, and nothing internal ever calls it except end.

At the possible expense of being "that guy," `.then()` uses [`.run()` internally too](https://github.com/segmentio/nightmare/blob/master/lib/nightmare.js#L357).

> I can see a lot of value in simply canonizing a thing lots of people seem to be using.

Adding some historical context:  I believe the PhantomJS implementation [used `.run()` to execute the Nightmare chain](https://github.com/segmentio/nightmare/tree/1.8.2#runcb).  I would hazard a guess that a fair number of things out there were either developed against 1.x or were developed looking at examples from 1.x.

> ...  so I think I lean towards deprecating run.

Making `.run()` a method that wraps an internal `run()` method with some deprecation warnings would be something I could get behind.  Thoughts?

Mr0grog:
> At the possible expense of being "that guy," .then() uses .run() internally too.

Oh, fine! ;)

> I believe the PhantomJS implementation used .run() to execute the Nightmare chain.

Ah! Well that explains a lot. That actually makes me feel not-so-hot about deprecating it, too :\

ripper234:
@rosshinkley writes in #979 

> TL;DR: don't use .run(). It's a holdover from Nightmare 1.x. Use .then() instead.

If that's indeed the case, let's decide this & close this issue.

My 2 cents, without getting into the whole history here, is the syntax of `run()` is nice, and I don't really understand why it can't/shouldn't be a wrapper. But whether it is supported or isn't supported, a decision one way or another would be nice :)

[...]

PR #900, opened 2016-11-20
`.end(fn)` should use `.then()`

### LINKED ISSUE #899: Passing a callback function to .end(someFunc) together with .evaluate freezes nightmare
Here's an example:

Working ( notice the .end() call )
```
nightmare
      .goto('http://yahoo.com')
      .wait(1000)
      .evaluate(() => 42)
      .end()
      .then((result) => {
        console.log(result); // Prints 42
      })
      .catch((error) => {
        console.error('Search failed:', error);
      });
```

NOT WORKING and freezing without doing anything
```
nightmare
      .goto('http://yahoo.com')
      .wait(1000)
      .evaluate(() => 42)
      .end(() => { console.log('code finished running'); })
      .then((result) => {
        console.log(result); // Statement never reached :(
      })
      .catch((error) => {
        console.error('Search failed:', error);
      });
```

I assume because some Promise is not chained properly, I have no idea why this is happening.

rosshinkley:
Fixes #899.

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -235,7 +235,15 @@ Set the `useragent` used by electron.
 Set the `user` and `password` for accessing a web page using basic authentication. Be sure to set it before calling `.goto(url)`.
 
 #### .end()
-Complete any queue operations, disconnect and close the electron process.  Note that if you're using promises, `.then()` must be called after `.end()` to run the `.end()` task.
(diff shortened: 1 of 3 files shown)


jekku:
Does this have breaking changes for anyone, boss?

justinmchase:
Why not just consider it an error case to pass a function into end? How is that better / different from just chaining thens after end?

rosshinkley:
@jekku  This probably should have gotten pulled in with the last release.  I'll see to it that it gets pulled in with the next one.

@justinmchase _mostly_ for previous version support/preventing breaking changes.  In the future, support could be dropped.  (This is _tangentially_ related to a conversation about `.run()` in #575.)

Issue #916, opened 2016-12-02
Nightmare should set a partition by default


justinmchase:
```    
this.browser = Nightmare({
  webPreferences: {
    partition: 'nightmare'
  }
});
```

I got burned pretty hard by the fact that I am starting and stopping the browser and there is some state that is getting stored between instances and it was affecting my tests on subsequent runs.

I would like to make a PR that automatically sets a partition by default. If a user wanted to persist state between tests they could set the partition to be `paritition: 'persist: whatever'` still to override it or `partition: null` to revert to the default.

Is there any reason why it would be good or should  actually be set to persist between browser runs? If not I will make a PR to change it. Otherwise I may just add to the front page documentation with a tip about setting the partition to prevent this issue.

overflowz:
Same problems here. I don't want sessions to be shared across different instances. I was trying to find the 'incognito' mode too, but no luck.

jekku:
I've been doing 'nopersist' all the time in my main config to avoid Nightmare from persisting states. I use it in integration tests where tests should be initially stateless and also independent from all other tests.

justinmchase:
@jekku Do you mean literally using the name `'nopersist'` as your partition name? I think you can use any name for the same effect but yeah that would work. I don't know of any other way to do it and you can get burned pretty hard if you don't.

PR #919, opened 2016-12-05
Alter default behavior to non-persistent partition

justinmchase:
The current behavior uses electron default, which creates a partition
for storage which persists between instances. This makes sense for
electron but not necessarily for primary use cases of Nightmare.

This change sets the default partition to be a non-persistent partition
but allows a user to opt back into persistence with simple configuration options.

Fixes segmentio/nightmare#916

## 4 new passing tests, all others green
screenshot 2016-12-05 16 59 20

## Preview of wiki changes
screenshot 2016-12-05 16 53 50

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -678,6 +678,34 @@ __nightmare.ipc = require('electron').ipcRenderer;
 
 To benefit of all of nightmare's feedback from the browser, you can instead copy the contents of nightmare's [preload script](lib/preload.js).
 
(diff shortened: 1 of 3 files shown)


rosshinkley:
More as a reminder to self when I pull this in: A note will need to be added to the release notes as a _possibly_ breaking change.  It's possible some folks out there are depending on this side effect, and it's probably worth mentioning.

Otherwise, looks great.  Thanks!

Issue #809, opened 2016-09-26
screenshot action hangs while pdf works fine


mikeys:
Hi guys,

Trying to leverage `nightmare` in order to be able to convert an svg to a png.
For some reason the following gets stuck on the `screenshot` action:

``` node
var Nightmare = require('nightmare')
var nightmare = Nightmare({show: true})
var url = "file:///Users/username/Pictures/whatever.svg"
var userAgent = "Mozilla/5.0 (Windows NT 6.3; WOW64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/38.0.2125.111 Safari/537.36"

nightmare
  .useragent(userAgent)
    .goto(url)
  .wait('svg')
  .evaluate(() => {
    const svgElem = document.querySelector('svg')

    return {
      width: svgElem.scrollWidth,
      height: svgElem.scrollHeight
    }
  })
  .then(function(dimensions) {
    return nightmare
      .viewport(dimensions.width, dimensions.height)
      .screenshot('./screenshot.png')
  })
  .then(function() {
    console.log('done')
    nightmare.end()
  })
  .catch(function(err) {
    console.log('error', err)
  })
```

Logs:

```
  nightmare queuing process start +0ms
  nightmare queueing action "useragent" +2ms
  nightmare queueing action "goto" for file:///Users/username/Pictures/whatever.svg +1ms
  nightmare queueing action "wait" +1ms
  nightmare queueing action "evaluate" +0ms
  nightmare running +1ms
  nightmare:actions .useragent() to Mozilla/5.0 (Windows NT 6.3; WOW64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/38.0.2125.111 Safari/537.36 +360ms
  nightmare:log Navigating: "file:///Users/username/Pictures/whatever.svg",
            headers: [none],
            timeout: 30000 +8ms
  nightmare:log did-start-loading +63ms [{}]
  nightmare:log did-get-response-details +77ms [{},true,"file:///Users/username/Pictures/whatever.svg","file:///Users/username/Pictures/whatever.svg",200,"GET","",{"access-control-allow-origin":["*"]},"mainFrame"]
  nightmare:log dom-ready +91ms [{}]
  nightmare:log did-frame-finish-load +0ms [{},true]
  nightmare:log did-finish-load +1ms [{}]
  nightmare:log did-stop-loading +1ms [{}]
  nightmare:actions .wait() for svg element +1ms
[...]

Issue #736, opened 2016-07-27
Nightmare freezes on either evaluate() or screenshot()


ryanvincent29:
I am working on moving from Phantom JS to Nightmare in an effort to get more CSS compatibility on my pages.

As a high-level view, my Nightmare instance lives inside a Docker container and listens for data to be pushed onto a queue. Once that data is pushed, I pull some data (including a string of HTML) out of the queue, alter the page a bit using Cheerio, and then turn the data into a HTML Data URL. 

Then, I goto the data url, use a couple of evaluate functions, scroll the page, and take a screenshot.

Here is a condensed version of my code:

``` javascript
function callNightmare(nightmare, html, data) {
  return new Promise(function(resolve, reject){
    var imageLoc = "./screenshot" + data.randomGuid + '.png';
    nightmare
      .userAgent(data.agentString)
      .goto(html)
      .wait()
      .then(function() {
        //remove scrollbar from image
        return nightmare.evaluate(function() { 
          var s = document.styleSheets[0];
          s.insertRule('::-webkit-scrollbar { display:none; }'); }
      })
      .then(function() {
        return nightmare.viewport(data.width, data.height);
      })
      .then(function() {
        return nightmare.inject('js', './localScript.js');
      })
      .then(function() {
        return nightmare.evaluate(getScrollPositionFromPage);
      })
      .then(function(scrollPosition) {
        return nightmare.scrollTo(scrollPosition.top, scrollPosition.left);
      })
      .then(function() {
        return nightmare.evaluate(fillInputsOnPageWithValues);
      })
      .then(function(){
        return nightmare.screenShot(imageLoc);
      })
      .then(function() {
        return nightmare.end();
      })
      .then(function() {
        resolve(imageLoc);
      })
      .catch(function(err) { 
        reject(err); 
      })
  });
}
```

However, my script will occasionally lock up during various evaluate functions, or during the screenshot function. Other times, the script runs successfully. I have tried this on various websites with the same results.

I can't pinpoint anything in the script that is causing this issue, and I didn't know if any of you had any thoughts on the problem?

rosshinkley:
Silly question: have you tried running with `DEBUG=*` to see if that has any helpful information?

rosshinkley:
(Also: Possibly related to #405.)

ryanvincent29:
Yeah, I run it with DEBUG=*, it outputs 2 lines indicating the initialization of Nightmare, but then I stop getting output from that debug path for the remainder of the program.

[...]

Issue #555, opened 2016-04-06
Screenshot action hangs application


bahlo:
Hey guys, thank you for this library, I'm really loving it (especially in comparison with the PhantomJS backend).

I'm creating screenshots of a page with multiple views with the following code.
The views are changed in the `nextView` function.

``` javascript
for (var i = 0; i < count; i++) {
  nightmare
    .wait(2000)
    .screenshot(path.resolve(__dirname, i + '.png'))
    .evaluate(nextView)
}

nightmare
  .end()
  .then(function() {
    console.log("Done")
  }).catch(function(err) {
    console.error(err);
  })
```

This works great on most pages, but it hangs on a specific page (I'm sorry I can't disclose the contents).
Here is a part of the log:

```
  nightmare:actions .evaluate() fn on the page +2ms
  nightmare:actions .wait() for 2000ms +6ms
  nightmare:actions .screenshot() +2s
  nightmare:log subscribing to browser window frames +1ms
  nightmare:log altering page to force rendering +0ms
  nightmare:log unsubscribing from browser window frames +57ms
  nightmare:actions .screenshot() captured with length 342104 +606ms
```

```
  nightmare:actions .evaluate() fn on the page +2ms
  nightmare:actions .wait() for 2000ms +392ms
  nightmare:actions .screenshot() +2s
  nightmare:log subscribing to browser window frames +0ms
  nightmare:log altering page to force rendering +0ms
```

As you can see, the screenshot before renders nicely, but the last one does not seem to do anything.
The [`requestFrame` function](https://github.com/segmentio/nightmare/blob/a215dd68fe3134d243a81af1723729474c0d4f85/lib/frame-manager.js#L52) doesn't seem to call its callback.

Do you have any tips for me to debug this? Setting a `waitTimeout` doesn't help since I'm not waiting. 

**Update:** Okay, I tried again with `show: yes` and manually triggered a re-render and it worked. Maybe I'll try to click on something to trigger that render manually every time.

**Update 2:** Since it's only waiting for change after the `screenshot` action, calling `scrollTo` or `click` before doesn't help.

**Update 3:** I tested it and the callback doesn't get called in the `requestFrame` function for this screenshot (all the other shots are working fine).

**Update  4:** I'm just calling the callback directly in `requestFrame` which is bad, but works for me. Gonna leave this issue open though.

bahlo:
[...]

PR #927, opened 2016-12-16
Use debugger highlighting for triggering frames to render (improves screenshots)

Mr0grog:
Since the initial implementation of `FrameManager`, Electron has added an API for Chromium's remote debugging protocol, which can visually highlight a portion of the page. This change uses that functionality to force a new frame to be rendered, which is much more reliable than the way Nightmare currently tries to force new frames to render by fiddling around with the DOM (see people experiencing problems with this in #555, #736, #809). It also has the benefit of not doing anything page content can observe, ensuring that any JS or CSS won't modify the page in response to Nightmare's attempt to take a screenshot.

This also adds a timeout. We should never hit it when using the debugger, but I figure it’s better to be safe than sorry here, especially after the hangs people have been experiencing.

In future versions of the remote debugging protocol, it will be possible to directly capture an image of the page, but that feature is still experimental (so it could be removed) and Electron does not yet support it anyhow. Something to keep in mind for future changes, though.

This is an alternative solution to the one in 53dee8a2ad88ea656d2b48504dbe66ff5287d0ab (currently on the `screenshot-with-offscreen-rendering` branch). That method (using Electron's new "offscreen" rendering mode) is *much* faster than this and vastly simplifies the code, but has more ways it can fail and has other side effects. If that approach looks better, I can submit it as a PR. We can also look into combining the two if the side effects are acceptable.

### DIFF
--- a/lib/frame-manager.js
+++ b/lib/frame-manager.js
@@ -61,7 +61,9 @@ function FrameManager(window) {
        many milliseconds, run the callback anyway. In this case, The
        callback's first argument, an image buffer, will be `null`.
    */
-  this.requestFrame = function(callback, timeout = 1000) {
+  this.requestFrame = function(callback, timeout) {
+    timeout = (timeout == undefined) ? 1000 : timeout;
+    
     if (callback) {
       this.once('data', callback);
     }
@@ -85,9 +87,9 @@ function FrameManager(window) {
       }
       
       if (timeout) {
-        frameRequestTimeout = setTimeout(() => {
+        frameRequestTimeout = setTimeout(function() {
           parent.emit('log', `FrameManager timing out after ${timeout} ms with no new rendered frames`);
-          this.emit('data', null)
+          self.emit('data', null)
         }, timeout);
       }
 
--- a/lib/frame-manager.js
+++ b/lib/frame-manager.js
@@ -98,6 +98,7 @@ function FrameManager(window) {
       window.webContents.debugger.sendCommand(
         'DOM.highlightRect', HIGHLIGHT_STYLE, function(error) {
           window.webContents.debugger.sendCommand('DOM.hideHighlight');
+          window.webContents.debugger.detach();
         });
     }
   };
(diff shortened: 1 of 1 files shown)


Mr0grog:
Note the offscreen rendering branch has window sizing issues as a result of an Electron bug (https://github.com/electron/electron/issues/8224), so it’s probably just untenable for now.

rosshinkley:
One question: if you're using the remote debugging protocol, doesn't that mean you can't use the debugger (or have the development tray) open at the same time?  If I'm not mistaken, this is the same issue that [`nightmare-upload`](https://github.com/rosshinkley/nightmare-upload) has.

It's also entirely possible this got fixed in a more recent version of Electron.  I guess what I'm beating around the bush and saying that either this needs to be tested for and/or should be included in the readme.  Otherwise, I'm +1 on this as it solves more problems than it creates, imho.

Thoughts?

Mr0grog:
> if you're using the remote debugging protocol, doesn't that mean you can't use the debugger (or have the development tray) open at the same time? If I'm not mistaken, this is the same issue that nightmare-upload has.

😲 I was entirely unaware of that as an issue and hadn’t tested it. The good news is that if a user has opened the local debugger before screenshotting, screenshots still work (see the `catch` clause here: https://github.com/segmentio/nightmare/pull/927/files#diff-bc4eacf94ad1a8e7c9bea8b6b6451251R83). The bad news is that users can’t open the local debugger *after* screenshotting. I’ve added another commit that detaches the debugger as soon as we’re done, which seems to resolve that.

Since you’ve seen this issue before, it’d be great if you give it a go as well, @rosshinkley.

rosshinkley:
> :astonished: I was entirely unaware of that as an issue and hadn’t tested it.

No worries.  I just knew this was a stumbling block in the past, thought I'd bring it up.

> I’ve added another commit that detaches the debugger as soon as we’re done, which seems to resolve that.

I re-read what you had done and tinkered for a bit, I'm going to go ahead and say LGTM and that this should get included.  Thanks!

Mr0grog:
Sweet! Glad this can improve things a little.

PR #929, opened 2016-12-19
Don't dispatch event on non-existent option.

notVitaliy:
In certain cases in angular, if there is a select box with an event listener that has no options populated and an event if dispatched, it will trigger change detection within angular, which will throw errors and cause electron to crash/freeze.

### DIFF
--- a/lib/actions.js
+++ b/lib/actions.js
@@ -311,11 +311,13 @@ exports.uncheck = function(selector, done){
 exports.select = function(selector, option, done) {
   debug('.select() ' + selector);
   this.evaluate_now(function(selector, option) {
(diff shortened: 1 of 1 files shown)


reinpk:
happy to consider if there's a test case and comments on an updated PR, thanks!

notVitaliy:
@reinpk What do you mean by test case? Like a unit-test? Or a codepen that shows the issue?

notVitaliy:
@reinpk This issue might be difficult to reproduce in a unit-test since it's an interaction between angular2 and nightmare, afaik. I'll do my best to isolate the underlying cause over the next couple weeks.

PR #930, opened 2016-12-20
Updated link to electron documentation

carlsa:
I have updated the link to the electron documentation to match the version that is specified in [package.json](https://github.com/segmentio/nightmare/blob/master/package.json) file.

This is relevant since the electron 1.4.4 documentation describes the option of specifying a custom page size when generating PDF-files.


reinpk:
thanks!

PR #865, opened 2016-11-02
Readme: update example to use duckduckgo.com

rstacruz:
Yahoo.com's home page took a very long time to work—for a moment I thought nightmare.js wasn't working!

nightmare

PR #941, opened 2017-01-02
Readme: make example actually work

rstacruz:
Seems like #865 took out these lines (sorry, my fault)


rosshinkley:
Thanks for the update.

PR #945, opened 2017-01-07
Repaired support for multiple timeouts in FrameManager

twolfson:
I'm a new user of Nightmare but found an issue when recording multiple screenshots on a static page (i.e. one with no frame updates). I was recording multiple screenshots in series at different viewport sizes.

- Nightmare was hanging indefinitely when hidden
- Nightmare was hanging when visible until I moved my mouse over the content (thus generating a new frame for the debugger)

After some sleuthing, it looks like FrameManager has some sneaky edge cases that needed deeper testing. In this PR:

- Fixed up double subscription issue due to still using `subscribed` boolean despite no longer being used
- Added `requestedFrame` resetting on `attach` error and `timeout` so future attach errors/timeouts can get the same `null` frame
- Added tests

**Notes:**

With respect to tests, I feel like I did a quick and dirty job. The `require` chunks aren't pulled out and I added `async` despite it not being used in the rest of the repo. I mostly wanted to get feedback on the PR and how y'all would like to handle the code setup instead of stressing myself out. I will gladly make any requested changes

Closes https://github.com/segmentio/nightmare/issues/555#issuecomment-324119322

### DIFF
--- a/lib/frame-manager.js
+++ b/lib/frame-manager.js
@@ -44,6 +44,12 @@ function FrameManager(window) {
     }
   }
 
+  function cancelFrame() {
+    requestedFrame = false;
+    clearTimeout(frameRequestTimeout);
+    self.emit('data', null);
+  }
+
   function receiveFrame(buffer) {
     requestedFrame = false;
     clearTimeout(frameRequestTimeout);
@@ -79,8 +85,7 @@ function FrameManager(window) {
         }
         catch (error) {
           parent.emit('log', `Failed to attach to debugger for frame subscriptions: ${error}`);
-          requestedFrame = false;
-          this.emit('data', null);
+          cancelFrame();
           return;
         }
       }
@@ -88,8 +93,7 @@ function FrameManager(window) {
       if (timeout) {
         frameRequestTimeout = setTimeout(function() {
           parent.emit('log', `FrameManager timing out after ${timeout} ms with no new rendered frames`);
-          requestedFrame = false;
-          self.emit('data', null)
+          cancelFrame();
         }, timeout);
       }
 
--- a/lib/frame-manager.js
+++ b/lib/frame-manager.js
@@ -79,6 +79,7 @@ function FrameManager(window) {
         }
         catch (error) {
           parent.emit('log', `Failed to attach to debugger for frame subscriptions: ${error}`);
+          requestedFrame = false;
           this.emit('data', null);
           return;
         }
@@ -87,6 +88,7 @@ function FrameManager(window) {
       if (timeout) {
         frameRequestTimeout = setTimeout(function() {
           parent.emit('log', `FrameManager timing out after ${timeout} ms with no new rendered frames`);
+          requestedFrame = false;
           self.emit('data', null)
         }, timeout);
       }
--- a/lib/frame-manager.js
+++ b/lib/frame-manager.js
@@ -23,7 +23,6 @@ function FrameManager(window) {
   if (!(this instanceof FrameManager)) return new FrameManager(window);
 
   EventEmitter.call(this);
-  var subscribed = false;
   var requestedFrame = false;
   var frameRequestTimeout;
   var self = this;
@@ -42,7 +41,6 @@ function FrameManager(window) {
     if (!self.listenerCount('data')) {
       parent.emit('log', 'unsubscribing from browser window frames')
       window.webContents.endFrameSubscription();
-      subscribed = false;
     }
   }
 
--- a/lib/frame-manager.js
+++ b/lib/frame-manager.js
@@ -32,7 +32,7 @@ function FrameManager(window) {
   this.on('removeListener', unsubscribe);
 
   function subscribe(eventName) {
-    if (!subscribed && eventName === 'data') {
+    if (!self.listenerCount('data') && eventName === 'data') {
       parent.emit('log', 'subscribing to browser window frames');
       window.webContents.beginFrameSubscription(receiveFrame);
     }
--- a/test/index.js
+++ b/test/index.js
@@ -1463,6 +1463,31 @@ describe('Nightmare', function () {
       didSubscribe.should.be.false;
     });
 
+    it.only('should subscribe to frames when requested necessary', function(done) {
+      var didSubscribe = false;
+      var didUnsubscribe = false;
+      var FrameManager = require('../lib/frame-manager.js');
+      var fn;
+      var manager = FrameManager({
+        webContents: {
+          debugger: {
+            isAttached: function() { return true; },
+            sendCommand: function(command) { if (command === 'DOM.highlightRect') { fn('mock-data'); }}
+          },
+          beginFrameSubscription: function(_fn) { didSubscribe = true; fn = _fn; },
+          endFrameSubscription: function() { didUnsubscribe = true; },
+          executeJavaScript: function() {}
+        }
+      });
+      manager.requestFrame(function (data) {
+        console.log('uhhh', data);
+        didSubscribe.should.be.true;
+        didUnsubscribe.should.be.true;
+        data.should.equal('mock-data');
+        done();
+      });
+    });
+
     it('should load jquery correctly', function*() {
       var loaded = yield nightmare
         .goto(fixture('rendering'))
--- a/test/index.js
+++ b/test/index.js
@@ -1564,7 +1564,7 @@ describe('Nightmare', function () {
     });
 
     // DEV: We can have multiple timeouts if page is static
-    it.only('should support multiple series timing out frame subscriptions', function(done) {
+    it('should support multiple series timing out frame subscriptions', function(done) {
       var subscribeCount = 0;
       var unsubscribeCount = 0;
       var FrameManager = require('../lib/frame-manager.js');
--- a/test/index.js
+++ b/test/index.js
@@ -1487,7 +1487,7 @@ describe('Nightmare', function () {
       });
     });
 
-    it.only('should support multiple concurrent frame subscriptions', function(done) {
+    it('should support multiple concurrent frame subscriptions', function(done) {
       var subscribeCount = 0;
       var unsubscribeCount = 0;
       var FrameManager = require('../lib/frame-manager.js');
@@ -1525,27 +1525,40 @@ describe('Nightmare', function () {
       });
     });
 
-    // TODO: Implement/test me
-    it.skip('should support multiple series frame subscriptions', function(done) {
+    it.only('should support multiple series frame subscriptions', function(done) {
       var subscribeCount = 0;
       var unsubscribeCount = 0;
       var FrameManager = require('../lib/frame-manager.js');
-      var fn;
+      var fn = null;
+      var assert = require('assert');
+      var async = require('async');
       var manager = FrameManager({
         webContents: {
           debugger: {
             isAttached: function() { return true; },
-            sendCommand: function(command) { if (command === 'DOM.highlightRect') { fn('mock-data'); }}
+            sendCommand: function(command) {
+              if (command === 'DOM.highlightRect') {
+                setTimeout(function () {
+                  fn('mock-data');
+                }, 100);
+              }
+            }
           },
-          beginFrameSubscription: function(_fn) { didSubscribe = true; fn = _fn; },
-          endFrameSubscription: function() { didUnsubscribe = true; },
+          beginFrameSubscription: function(_fn) { subscribeCount += 1; assert.strictEqual(fn, null); fn = _fn; },
+          endFrameSubscription: function() { unsubscribeCount += 1; fn = null; },
           executeJavaScript: function() {}
         }
       });
-      manager.requestFrame(function (data) {
-        didSubscribe.should.be.true;
-        didUnsubscribe.should.be.true;
-        data.should.equal('mock-data');
+      async.timesSeries(2, function requestFrameFn (i, cb) {
+        manager.requestFrame(function handleFrame (data) {
+          cb(null, data);
+        });
+      }, function handleResults (err, results) {
+        if (err) { done(err); }
+        subscribeCount.should.equal(2);
+        unsubscribeCount.should.equal(2);
+        results[0].should.equal('mock-data');
+        results[1].should.equal('mock-data');
         done();
       });
     });
--- a/test/index.js
+++ b/test/index.js
@@ -1525,7 +1525,7 @@ describe('Nightmare', function () {
       });
     });
 
-    it.only('should support multiple series frame subscriptions', function(done) {
+    it('should support multiple series frame subscriptions', function(done) {
       var subscribeCount = 0;
       var unsubscribeCount = 0;
       var FrameManager = require('../lib/frame-manager.js');
--- a/test/index.js
+++ b/test/index.js
@@ -1563,6 +1563,39 @@ describe('Nightmare', function () {
       });
     });
 
+    // DEV: We can have multiple timeouts if page is static
+    it.only('should support multiple series timing out frame subscriptions', function(done) {
+      var subscribeCount = 0;
+      var unsubscribeCount = 0;
+      var FrameManager = require('../lib/frame-manager.js');
+      var fn = null;
+      var assert = require('assert');
+      var async = require('async');
+      var manager = FrameManager({
+        webContents: {
+          debugger: {
+            isAttached: function() { return true; },
+            sendCommand: function() { /* Ignore command so it times out */ }
+          },
+          beginFrameSubscription: function(_fn) { subscribeCount += 1; assert.strictEqual(fn, null); fn = _fn; },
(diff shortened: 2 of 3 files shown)


test/index.js
```diff
@@ -1463,6 +1463,139 @@ describe('Nightmare', function () {
...
+        if (err) { done(err); }
+        subscribeCount.should.equal(2);
+        unsubscribeCount.should.equal(2);
+        results[0].should.equal('mock-data');
+        results[1].should.equal('mock-data');
+        done();
+      });
+    });
+
+    // DEV: We can have multiple timeouts if page is static
```
Mr0grog: For what it’s worth, the entire point of #927 was to not have timeouts when the page is static ;)
(The timeout guard should really never fire at all, but is there just in case the debugger fails in ways it should probably not be able to or its API has a breaking change.)

test/index.js
```diff
@@ -1463,6 +1463,139 @@ describe('Nightmare', function () {
...
+            isAttached: function() { return true; },
+            sendCommand: function(command) {
+              if (command === 'DOM.highlightRect') {
+                setTimeout(function () {
+                  fn('mock-data');
+                }, 100);
+              }
+            }
+          },
+          beginFrameSubscription: function(_fn) { subscribeCount += 1; assert.strictEqual(fn, null); fn = _fn; },
```
Mr0grog: Is there a particular reason to use `assert` here instead of Chai? e.g. `should.not.exist(fn);`

lib/frame-manager.js
```diff
@@ -32,7 +31,7 @@ function FrameManager(window) {
   this.on('removeListener', unsubscribe);
 
   function subscribe(eventName) {
-    if (!subscribed && eventName === 'data') {
+    if (!self.listenerCount('data') && eventName === 'data') {
```
Mr0grog: Oh, geez, I really lost track of cleaning up that `subscribed` flag! This approach is way better 👍

Mr0grog: Oh, geez, I really lost track of cleaning up that `subscribed` flag! This approach is way better 👍

test/index.js
```diff
@@ -1463,6 +1463,139 @@ describe('Nightmare', function () {
       didSubscribe.should.be.false;
     });
 
+    it('should subscribe to frames when requested necessary', function(done) {
+      var didSubscribe = false;
+      var didUnsubscribe = false;
+      var FrameManager = require('../lib/frame-manager.js');
```
Mr0grog: Same, obviously for the rest of these (and the other `require` calls) in the other 3 tests you added.

test/index.js
Mr0grog: Relatedly, I might re-title this test to something like “should time out with null data when debugger fails.”

lib/frame-manager.js
```diff
@@ -42,10 +41,15 @@ function FrameManager(window) {
...
-      subscribed = false;
     }
   }
 
+  function cancelFrame() {
+    requestedFrame = false;
+    clearTimeout(frameRequestTimeout);
+    self.emit('data', null);
+  }
+
```
twolfson: Ah, good insight. I didn't even see that o_o

If we are going for conciseness, we could probably do 1 better and remove `cancelFrame()` altogether (i.e. call `receiveFrame(null);` at those locations).

If we are going for clarity (my preference), I sometimes will sacrifice DRYness for making the intent clear (i.e. we mean to cancel the frame request and emit `null` data). In the concise scenario and another contributor reading the code, they might be uncertain if we actually wanted to emit `null` or not =/

test/index.js
```diff
@@ -1463,6 +1463,139 @@ describe('Nightmare', function () {
...
+        webContents: {
+          debugger: {
+            isAttached: function() { return true; },
+            sendCommand: function(command) { if (command === 'DOM.highlightRect') { fn('mock-data'); }}
+          },
+          beginFrameSubscription: function(_fn) { didSubscribe = true; fn = _fn; },
+          endFrameSubscription: function() { didUnsubscribe = true; },
+          executeJavaScript: function() {}
+        }
+      });
```
twolfson: I'm on the fence about this. I like the idea but...

- We define custom hooks in each of the tests so it won't save us too many lines of code. It might even lead to more confusion due to jumping from code to definition and back. That is:

```
var manager = FrameManager({
  webContents: {
    debugger: {
      isAttached: function() { return true; },
      sendCommand: function(command) { if (command === 'DOM.highlightRect') { fn('mock-data'); }}
    },
    beginFrameSubscription: function(_fn) { didSubscribe = true; fn = _fn; },
    endFrameSubscription: function() { didUnsubscribe = true; },
    executeJavaScript: function() {}
  }
});
```

will become:

```
var mockWindow = MockWindow({
  // Unsure about how we define `fn` access at the moment
  sendCommand: function(command) { if (command === 'DOM.highlightRect') { this.fn('mock-data'); }}
  beginFrameSubscription: function() { didSubscribe = true; },
  endFrameSubscription: function() { didUnsubscribe = true; }
})
```

- This file is quite lengthy and isn't broken down, where would the class definition live? My gut says top of the file or in a separate file but both feel a little premature/less clear when reading code

test/index.js
twolfson: There is! =D I prefer to use `assert` like a developer sanity check and `should/expect` for test assertions. It's kind of like `undefined` and `null` (i.e. one is for the system, one is for programmers)

I'm glad to change this based on the repo's coding styles but I didn't see any programmatic `assert` sanity checks when browsing =/

test/index.js
twolfson: Ah, interesting. I was getting consistent timeouts hence my PR in the first place o_o The script can be found here:

https://github.com/twolfson/multi-image-mergetool/blob/1.32.1/bin/_build-demo-screenshots.js#L75-L164

twolfson:
Thanks for the feedback @Mr0grog =) I left :+1: on comments I will definitely take care of and feedback on other ones for longer discussions. I'm going to wait until repo owners leave comments until I take care of the :+1:'d comments though (saves me time with less context switching)

lib/frame-manager.js
Mr0grog: Yeah, that’s why I gave two solutions :P   I think I agree with you that clarity is better, so I’d keep `cancelFrame()` around but rewrite it to just delegate to `receiveFrame(null)`.

test/index.js
Mr0grog: Oh, that’s not what I was trying to suggest at all. I was thinking you might have something roughly like…

```js
function MockFrameManagerWindow (options) {
  options = Object.assign({
    debuggerAlreadyAttached: false,
    debuggerTriggersFrames: true
  }, options);

  var self = this;
  var _debuggerAttached = !!options.debuggerAlreadyAttached;
  var _frameCallback = function() {};

  self.results = {
    subscribedCount: 0,
    unsubscribedCount: 0,
    currentSubscriptionCount: 0
  };

  self.beginFrameSubscription = function (callback) {
    _frameCallback = callback;
    self.results.subscribedCount++;
    self.results.currentSubscriptionCount++;
  };

  self.endFrameSubscription = function () {
    _frameCallback = function() {};
    self.results.unsubscribedCount++;
    self.results.currentSubscriptionCount--;
  };

  self.debugger = {
    isAttached: function () { return _debuggerAttached; },
    attach: function () {
      asset(!_debuggerAttached, 'Debugger is already attached');
      _debuggerAttached = true;
    },
    detach: function () {
      _debuggerAttached = false;
    },
    sendCommand: function (command, callback) {
      if (options.debuggerTriggersFrames && command === 'DOM.highlightRect') {
        setTimeout(_frameCallback, 0, 'mock_data');
      }
      if (callback) {
        setTimeout(callback, 0);
      }
    }
  };
}
```

If you wanted to do less work here, you could use [Sinon](http://sinonjs.org) to make all the functions; it automatically keeps tracked of whether they were called and how often.

Anyway, instead of using local closure variables, you’d just test `mockInstance.results.whatever` to find out whether things were called or not.

lib/frame-manager.js
twolfson: Oh, I totally missed that first part. That's what I get for reading it half-awake x_x

test/index.js
Mr0grog: > I'm glad to change this based on the repo's coding styles but I didn't see any programmatic assert sanity checks when browsing

There aren’t. I generally like to do that, too, but I my perception on this is that any assertion that only exists in the context of a test is a test assertion, not an “is the program in a good state” assertion. But I’m definitely not in charge of the coding style here *at all.*

test/index.js
twolfson: Ahh, so we could stub the mock's method for causing timeouts. I dig it :+1: 

I typically use live data instead of mocks but that's a neat technique =)

twolfson: Fwiw, we don't need `currentSubscriptionCount` but if we did, it would always be `subscribedCount - unsubscribedCount` =/

Mr0grog: > This file is quite lengthy and isn't broken down, where would the class definition live?

Yeah, I am also reminded of how painfully long the test file is whenever I look at it—I think it’s a consequence of it having grown pretty organically and the Segment team not seeming to put much effort into this project anymore.

In the past, I’ve added support items to the bottom where there are a few such things (see [this custom Chai assertion, for example](https://github.com/segmentio/nightmare/blob/master/test/index.js#L2368-L2378)), but I think a mock class like I’m suggesting would be too big for that. I’d either:
- Just put it in a separate file in the same directory (e.g. how `server.js` is)
- Create a `test/support` directory and put it in there

test/index.js
Mr0grog: Well, that seems very bad. :|

test/index.js
twolfson: Hmm, I would say it's on the fence. On one hand, it's faking out the `beginFrameSubscription` to prevent it's `receiveFrame` multiple times (i.e. programmatic). On the other hand, it's verifying that our `async.times` doesn't invoke the `beginFrameSubscription` multiple times (i.e. assertion)

test/index.js
twolfson: Maybe we could break out the frame manager tests into their own file altogether? It's only 1 test prior this this, it would eliminate worries, and it would more towards a cleaner repo (i.e. 1 test file per 1 repo file)

test/index.js
Mr0grog: > On one hand, it's faking out the beginFrameSubscription to prevent it's receiveFrame multiple times (i.e. programmatic). On the other hand, it's verifying that our async.times doesn't invoke the beginFrameSubscription multiple times (i.e. assertion)

Wellllllllll, in the sense that it’s mocking Electron’s behavior, Electron allows `beginFrameSubscription` to be called multiple times—it just replaces the existing subscriber with no checks: https://github.com/electron/libchromiumcontent/blob/2249b37b0c3cb2e7ab4e333fbd3f87356beba55f/patches/guest_view.patch#L33-L36

I do think it’s totally reasonable for the test to be checking that we never call that method unless we’ve never subscribed or have already ended the subscription. In my book, that’s more of a test assertion.

But either way, I don’t think it’s a huge deal.

test/index.js
Mr0grog: > Maybe we could break out the frame manager tests into their own file altogether?

Sure. To be painfully honest, I don’t think I really have enough personal investment in Nightmare to push for reorganizing the repo. I made some noise to this effect way back near the start of 2016 and felt like I was mostly talking to a brick wall with the repo owners. You might have a different/better experience now, though.

test/index.js
Mr0grog: @twolfson do you have a more reduced test case than the script you linked to that demonstrates timeouts?

test/index.js
twolfson: Fair enough. I guess we'll wait until we get feedback from an owner =/

test/index.js
twolfson: Unfortunately no =/ I did verify that these test cases reproduced the error and when they were patched, so was the error

twolfson:
@Mr0grog On a tangent, did we ever explore using `requestAnimationFrame` instead of trying to force a DOM redraw?

Mr0grog:
> @Mr0grog On a tangent, did we ever explore using requestAnimationFrame instead of trying to force a DOM redraw?

Yes. It doesn’t trigger anything—it only schedules things for the next time a frame is prepared (before drawing).

twolfson:
Ah, alright. Worth asking about. Thanks =)

test/index.js
Mr0grog: Oh, ok, so they didn’t time out any more then? That would seem to indicate that the issue was the broken flag, not an actual timeout on rendering the frame. Phew.

Mr0grog:
Haha, no worries, I would have definitely asked the same question if I were in your position :)

test/index.js
twolfson: Nope, they still timeout but now timeouts in series actually call `setTimeout` instead of being stopped early by an un-reset `requestedFrame`

Mr0grog: Can you file a separate issue, then? And include what version of Electron you are using, since that affects the debugger API?

twolfson: Sure thing, I'm not in a place where I have the details but I'll do that later and `cc` you on it :+1:

Mr0grog: Thanks! No rush; I’m just about certain I won’t have time to look at it today anyway.

twolfson: Fwiw, my guess is `electron@1.4.13` as I was wrote that script 2/3 days ago and there hasn't been a new `electron` release in 20 days

https://github.com/electron/electron/releases

https://github.com/segmentio/nightmare/blob/2.9.1/package.json#L23

twolfson: Finally got around to making a minimal version for you. Sorry about the delay, I accidentally archived the email and then forgot about it until today. Here's the issue:

https://github.com/segmentio/nightmare/issues/955

Mr0grog:
The fix for two other issues (#955, #964) depends on the changes here and is also in the same area of the codebase. There’s an outline for what needs to change in #955, but I don’t really want to do that work until this PR is mostly set to go because they *will* conflict.

cscanlin:
I'm having this an issue with nightmare 2.10.0 that I believe is related to this PR.

Specifically for taking screenshots with varying viewport sizes:

Here's a debug log on a crashed run:

```
  nightmare:log did-frame-finish-load [{},true] +4ms
  nightmare:log did-finish-load [{}] +0ms
  nightmare:log page-favicon-updated [{},["http://cscanlin.github.io/periodic-table-timeline/beaker.ico"]] +0ms
  nightmare:log did-stop-loading [{}] +1ms
  nightmare:actions .viewport() +0ms
  nightmare:actions .screenshot() +57ms
  nightmare:log subscribing to browser window frames +1ms
```

And here's a debug log when it finished successfully:

 ```
  nightmare:log did-frame-finish-load [{},true] +28ms
  nightmare:log did-finish-load [{}] +1ms
  nightmare:log page-favicon-updated [{},["http://cscanlin.github.io/periodic-table-timeline/beaker.ico"]] +0ms
  nightmare:log did-stop-loading [{}] +0ms
  nightmare:actions .viewport() +1ms
  nightmare:actions .screenshot() +40ms
  nightmare:log subscribing to browser window frames +1ms
  nightmare:log Highlighting page to trigger rendering. +0ms
  nightmare:log unsubscribing from browser window frames +108ms
  nightmare:actions .screenshot() captured with length 104999 +296ms
```

Will this PR fix my issue? If so, and idea when it might be merged?

Thanks in advance, and let me know if I can provide more detail.

twolfson:
It looks similar to my problem so I would bet it would fix it. Since this PR hasn't been touched in 2 months and as @Mr0grog has previously suggested, we're probably going to be waiting for a while.

I suggest working around it in the meantime by pointing to a fork. This can be done via `git` URLs in your `package.json`:

https://docs.npmjs.com/files/package.json#git-urls-as-dependencies

https://github.com/twolfson/multi-image-mergetool/blob/1.32.1/package.json#L90

cscanlin:
@twolfson That did the trick, thanks! On that note, I was also able to merge your branch with the master branch with 0 conflicts. Are there any other major roadblocks in getting this deployed? @Mr0grog, you said that this branch will actually open up the path for even more fixes (in addition to fixing this issue), so it seems like an easy win. Thoughts?

Mr0grog:
@cscanlin I don’t think anybody who has commented here has any commit rights, so I dunno. It doesn’t seem like there should be any roadblocks here and I would *think* this should be high priority since it’s really easy to trigger it and hang Nightmare.

_Paging @rosshinkley…_

ledniy:
So why it's not merged? Without it screenshot doesn't work properly

cianBuckley:
could this be merged? currently creating my own nightmare version to patch this...

keithkml:
Are there any plans to cut a release of Nightmare to include this fix? Or should I start building my code against master?

nettofarah:
cc @fouad

PR #950, opened 2017-01-09
Add jpeg support

kennuzzo:
Hi, ive added support for a toJPEG() to nightmare since electron supports it too.
for consistency ive added a toPNG(), while the current `screenshot()` is an alias of it.

runner.js:
added toJPEG

Action.js
* added toJPEG
* alias screenshot with toPNG
* Theres another change on the `toPNG` where if clip is a function, i check if the path is an object instead of checking if its a string when i am assigning to a clip, so if i accidentally pass a number like a do to a `toJPEG` for the compression the value will get ignored.

Thank you

### DIFF
--- a/test/index.js
+++ b/test/index.js
@@ -1393,7 +1393,7 @@ describe('Nightmare', function () {
       var stats = fs.statSync(tmp_dir+'/test.jpg');
       Buffer.isBuffer(image).should.be.true;
       var mime = fileType(image).mime;
-      mime.should.equal("image/jpeg");
+      mime.should.equal('image/jpeg');
       image.length.should.be.at.least(1000);
     });
     it('should take a clipped jpeg screenshot', function*() {
@@ -1436,9 +1436,9 @@ describe('Nightmare', function () {
         });
       Buffer.isBuffer(image).should.be.true;
       var statsClipped = fs.statSync(tmp_dir+'/test-clipped.jpg');
-      image.length.should.be.eql(statsClipped.size);
+      image.length.should.be.equal(statsClipped.size);
       var mime = fileType(image).mime;
-      mime.should.equal("image/jpeg");
+      mime.should.equal('image/jpeg');
       image.length.should.be.at.least(300);
     });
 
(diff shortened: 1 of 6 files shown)


jimmaay:
Looks good, I would follow #135 though and just parse the file path and add an optional parameter for JPEG output quality (0-100). If no file extension then default to PNG for full backwards compatibility.

kennuzzo:
As suggested, moved back everything into `.screenshot()` which behave the same for `png` and `jpeg`,  similar to the `png` `.screenshot()` if we want the png version, we need to pass a compression `.screenshot(90)`.
Also added mime check on the tests to make sure that the buffer is an `image/jpeg` or `image/png`

test/index.js
```diff
@@ -1372,6 +1374,76 @@ describe('Nightmare', function () {
...
+          x: 200,
+          y: 100,
+          width: 100,
+          height: 100
+        });
+      Buffer.isBuffer(image).should.be.true;
+      var statsClipped = fs.statSync(tmp_dir+'/test-clipped.jpg');
+      image.length.should.be.eql(statsClipped.size);
+      var mime = fileType(image).mime;
+      mime.should.equal("image/jpeg");
```
briwa: I would rather use single quotes throughout all the codes, for consistency hehehe

test/index.js
```diff
@@ -1372,6 +1374,76 @@ describe('Nightmare', function () {
...
+        .goto('https://github.com')
+        .screenshot(90,{
+          x: 200,
+          y: 100,
+          width: 100,
+          height: 100
+        });
+      Buffer.isBuffer(image).should.be.true;
+      var statsClipped = fs.statSync(tmp_dir+'/test-clipped.jpg');
+      image.length.should.be.eql(statsClipped.size);
```
briwa: Maybe this one can also use `should.equal` for consistency? Or is there any specific reason of doing it this way?

casesandberg:
Can you update this to make sure its passing all the tests?

matthewmueller:
Going to close this since this PR is stale. Thanks for your efforts!

PR #967, opened 2017-01-22
Make nightmare isomorphic

### LINKED ISSUE #393: Support for running inside a normal electron app
Currently `nightmare` assumes that it will be running inside a normal `node.js` app so it ends using `child_process.spawn(/path/to/electron` to do its job, what i'm proposing is support a way to execute `nightmare` inside a native electron app, without spawning anything just using it like a normal module.

this feature will let `nightmare` to be running with a workers manager module (something like [electron-workers](https://github.com/bjrmatos/electron-workers)) and do tasks in scale.

if you are ok i can implement this on my own and send a PR

what i would like is to discuss with you guys is how would you like to support this feature.

some options:

- a new option to the nigthmare constructor? (`Nightmare({ nativeElectronApp: true })`)
- detecting if `nightmare` is running inside `electron` via a run-time condition

Timer:
This allows nightmare to run under an already executing electron application.
Resolves #393.

### DIFF
--- a/lib/ipc.js
+++ b/lib/ipc.js
@@ -32,6 +32,8 @@ function IPC(process) {
   }
 
   var emitter = process[instance] = new Emitter();
+  emitter.process = process
+  emitter._IS_IPC = true
   var emit = emitter.emit;
   var callId = 0;
   var responders = {};
--- a/lib/nightmare.js
+++ b/lib/nightmare.js
@@ -25,6 +25,7 @@ var split2 = require('split2');
 var defaults = require('defaults');
 var noop = function() {};
 var keys = Object.keys;
+const EventEmitter = require('events');
 
 // Standard timeout for loading URLs
 const DEFAULT_GOTO_TIMEOUT = 30 * 1000;
@@ -47,7 +48,12 @@ const DEFAULT_PARTITION = 'nightmare';
  * Export `Nightmare`
  */
 
+ const shouldMock = (function() {
+   return process.env._.endsWith('electron')
+ })()
+
 module.exports = Nightmare;
+module.exports.shouldMock = shouldMock
 
 /**
  * runner script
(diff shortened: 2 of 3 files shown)


jekku:
How do we test for isomorphism? Is it possible to write a test for this?

Timer:
@jekku I'm sure it's possible to test for isomorphism -- we'd just need to run the tests under electron instead of node.

Timer:
Unfortunately, I don't have time right now to explore adding tests for this.

I'm currently working on an application where time-to-market is critical and I needed nightmare to work under electron. This PR works for me and I have not explored all the caveats of using this method, but we can explore those at a later date.

This PR was mainly opened for those individuals following #393 and who need a solution before segmentio explores official support for running under `electron`.

FrantisekGazo:
@Timer please, how did you make it work on electron app? do you have some public electron app, that uses your fork?

I tried something like this:
```JavaScript
const Nightmare = require('nightmare');
const nightmare = Nightmare({
    gotoTimeout: 1000, // in ms
    loadTimeout: 1000, // in ms
    executionTimeout: 1000, // in ms
});
function run() {
    console.log('test start');

    const n = nightmare.goto('http://www.google.com')
        .end()
        .then(function (result) {
            console.log('test done', result);
        })
        .catch(function (error) {
            console.error('test failed:', error);
        });

    console.log('test running', n);
    return n;
}
```
It returns a Promise and it never resolves. I just get 2 outputs:
```
test start
test running Promise {[[PromiseStatus]]: "pending", [[PromiseValue]]: undefined}
```

Do I have to setup something more?

Timer:
@FrantisekGazo it works for me with just that patch, are you running it from the main process? It needs to be on the main process not the renderer.

Alternatively you can try my published version (which I depend on in prod), `eramthgin@2.9.3` and make sure you switch your require from `nightmare` to `eramthgin`.

FrantisekGazo:
yes, that was the problem - I ran it from a renderer process.
thanks @Timer 

But in the end I managed to do what I needed with [xray](https://github.com/lapwinglabs/x-ray).

Timer:
Ah, neat. I need javascript etc to be executed so I can't switch. :(

marcelfalliere:
Hey @Timer 

I'am using your module `eramthgin`, I think I'm in the same case as you... works great.

My question is do you see a flicker on your main window ? Even with `show:false` ... there's clearly a window that show and immediatly hides. Don't know if it's a dev-env problem, if this window is the dev-tools for the nightmare window ... don't know! Do you see something ? Or maybe you use it with show:true ?

Timer:
I use show false but I've never seen a flicker. I'm not sure what's causing that.

marcelfalliere:
It opens up the devtools for a brief time ... like for 800ms or so, the time for the nightmare.js to do its business. 

How I know ? Made a screen recording, and going image by image I see something this :
screen shot 2017-02-24 at 10 25 59
- background is the console
- the green on the right is the actuall electron app main window
- the already open devtools are the one for this main window
- the status bar in the middle is displayed and hidden in a flash ... voila

Will check if it also pops up when building the app.

lib/ipc.js
```diff
@@ -32,6 +32,8 @@ function IPC(process) {
   }
 
   var emitter = process[instance] = new Emitter();
+  emitter.process = process
+  emitter._IS_IPC = true
```
nick-bogdanov: missing `;`

matthewmueller:
Going to close this since this PR is stale. This is a great first step and I'd like this to be the next feature we add to nightmare. Thanks for your efforts!

Issue #981, opened 2017-01-28
Exception while running concurrently Nightmares


ripper234:
I'm getting the following exception:

```
  1) MyTest should work concurrently:
     Uncaught TypeError: Cannot read property 'apply' of undefined
      at Nightmare.done (node_modules\nightmare\lib\nightmare.js:353:14)
      at Nightmare.next (node_modules\nightmare\lib\nightmare.js:329:35)
      at self.child.call (node_modules\nightmare\lib\nightmare.js:341:46)
      at EventEmitter.<anonymous> (node_modules\nightmare\lib\ipc.js:93:18)
      at ChildProcess.<anonymous> (node_modules\nightmare\lib\ipc.js:49:10)
      at process.nextTick (internal/child_process.js:744:12)
      at _combinedTickCallback (internal/process/next_tick.js:67:7)
      at process._tickCallback (internal/process/next_tick.js:98:9)
```

When running the following concurrent test (The game is built so it only starts when 3 concurrent users login):

```
function loginPlayer(username) {
  return new Nightmare()
    .goto('http://localhost:8080')
    .type('#choosenickname', username)
    .click('#write-btn')
    .wait('div.game')
}

describe('MyTest', function () {
  this.timeout(20000);
  it('should work concurrently', function (done) {
    loginPlayer('1stuser').run()
    loginPlayer('2nduser').run()
    loginPlayer('3rduser').run(function (err, result) {
      done();
    })
  })
});
```

When I observe the behavior on Atom using `show: true`, I see that indeed 3 browsers open, login with different usernames, and everything is as expected (except the test failing).

noahtkeller:
This is occurring because it is attempting to execute the callback provided in the run block without a safety check. For a temporary work-around you can pass blank callbacks and it should work.

```
describe('MyTest', function () {
  this.timeout(20000);
  it('should work concurrently', function (done) {
    var finished = 0;
    function cb(err, result) {
      if (++finished === 3) {
        done();
      }
    }
    loginPlayer('1stuser').run(cb)
    loginPlayer('2nduser').run(cb)
    loginPlayer('3rduser').run(cb)
  })
});
```

ripper234:
Awesome, thanks for the quick workaround + fix.

PR #982, opened 2017-01-28
Closes segmentio/nightmare#981 - Exception while running concurrently…

noahtkeller:
… Nightmares

modified:   lib/nightmare.js

* Check typeof callback and set blank function if needed


casesandberg:

PR #983, opened 2017-01-28
Updated examples to es6

amilajack:


Readme.md
```diff
@@ -30,52 +30,48 @@ Many thanks to [@matthewmueller](https://github.com/matthewmueller) and [@rosshi
 Let's search on DuckDuckGo:
 
 ```js
-var Nightmare = require('nightmare');		
-var nightmare = Nightmare({ show: true });
+const Nightmare = require('nightmare');
```
jekku: I don't see much of 'require' in ES6 production code

amilajack: Would you recommend es6 imports?

jekku: I would probably say ES5 is a subset of ES6 since it's backwards compatible. But for this to be fully ES6, a destructuring import or es6 import would look better.

jekku:
I'm still a little on edge about this one. There will be a difference between

```js
import {default as Nightmare} from 'nightmare';
```

and

```js
import Nightmare from 'nightmare';
```

due to changes introduced by the latest versions of babel about the default import.

Kindly double check which one of these would work with the example, the one we are presenting here could or could not be the faulty one.

stephenmathieson:
@jekku Babel inserts a function to handle this for us:

```js
function _interopRequireDefault(obj) { return obj && obj.__esModule ? obj : { default: obj }; }
```

stephenmathieson:
I do not see the benefit in this change. It seems like unnecessary cosmetic stuff, as functionally every example code block will work exactly the same.

jekku:
@stephenmathieson thanks for teaching me that. :)

For me though, there could be some benefit on the es6 examples change - but it has subtle drawbacks as well. 

For people who need to be convinced to use this open source repository, it should be 'plug and play' as much as possible. Since not all ES6 features are available yet on the native engine without transpiling it from ES6 to ES5, they'd still add a layer of configuration above it to allow this example. I think they could figure out the cosmetic code changes on their own AFTER they are convinced to use nightmare.js.

casesandberg:
I feel like ES6 has some broad support now, I am going to go ahead and merge this.

PR #988, opened 2017-02-01
Mention mocha default timeout on README

sanbor:
Almost for sure users that want to try the example test snippet are going to run into issues with the default mocha timeout in the CLI

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -82,6 +82,7 @@ describe('test duckduckgo search results', function() {
   });
 });
 ```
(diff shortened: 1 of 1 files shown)


jekku:
@sanbor that is not the only case where test engineers should increase the mocha timeout for nightmare.js. It could also be possible that the operations being done on the browser are intentionally taking long.

sanbor:
@jekku Right. Does that means that you think that it's a good idea to mention this but with a different wording? Thanks.

jekku:
@sanbor yes

casesandberg:

PR #992, opened 2017-02-06
Override nightmare `Promise` library

itsjustcon:
Native ES6 Promises are nice, but libraries like [Bluebird](npmjs.com/package/bluebird) offer a lot of nice additional features. I'd like to be able to specify which Promise library/factory nightmare uses internally to wrap return values from `.then()` and `.catch()` *(inspired by [mongoose.Promise](http://mongoosejs.com/docs/promises.html#plugging-in-your-own-promises-library))*

## Use Case
```js
import Nightmare from 'nightmare';
import Promise from 'bluebird';

class MyCrawler {

  nightmare = null;

  // PRIVATE
  _authPromise = null;

  constructor() {
    this.nightmare = Nightmare();
  }

  async authenticate() {
    if (this._authPromise && !this._authPromise.isFulfilled()) {
      return await this._authPromise();
    } else {
      return await (this._authPromise = Promise.resolve(this.nightmare.goto('...').then()));
    }
  }

}
```

The problem here is the `Promise#isFulfilled()` method that is a Bluebird-only method, forcing me to wrap `Nightmare#then()` in an awkward `Promise.resolve(...)`.

## Solution
With the ability to specify the internal `Promise` library/factory, I could have nightmare return a Bluebird Promise:
```js
import Nightmare from 'nightmare';
import Promise from 'bluebird';

Nightmare.Promise = Promise; // Option #1

class MyCrawler {

  nightmare = null;

  // PRIVATE
  _authPromise = null;

  constructor() {
    this.nightmare = Nightmare({ Promise: Promise }); // Option #2
    this.nightmare = Nightmare({ Promise }); // Option #3 (b/c ES6 is AWESOME!)
  }

  async authenticate() {
    if (this._authPromise && !this._authPromise.isFulfilled()) {
      return await this._authPromise();
    } else {
      return await (this._authPromise = this.nightmare.goto('...').then());
    }
  }

}
```

Please review my code & consider it for inclusion in the next release. `nightmare` is one of my favorite & most-used NPM packages, and I think this feature would be a really, really useful addition!

### DIFF
--- a/test/index.js
+++ b/test/index.js
@@ -1892,7 +1892,61 @@ describe('Nightmare', function () {
       nightmare = Nightmare({ electronPath: require('electron') });
       nightmare.should.be.ok;
     });
+
+    it('should allow to use external Promise', function*() {
+      nightmare = Nightmare({ Promise: require('bluebird') });
+      nightmare.should.be.ok;
+      var thenPromise = nightmare.goto('about:blank').then();
+      thenPromise.should.be.an.instanceof(require('bluebird'));
+      yield thenPromise;
+      var catchPromise = nightmare.goto('about:blank').catch();
+      catchPromise.should.be.an.instanceof(require('bluebird'));
+      yield catchPromise;
+      var endPromise = nightmare.goto('about:blank').end().then();
+      endPromise.constructor.should.equal(require('bluebird'));
+      endPromise.should.be.an.instanceof(require('bluebird'));
+      yield endPromise;
+    });
   });
+
+  describe('Nightmare.Promise', function() {
+    var nightmare;
+    afterEach(function*() {
+      // `withDeprecationTracking()` messes w/ prototype constructor references
+      Nightmare.Promise = require('..').Promise = Promise;
+      yield nightmare.end();
+    });
+
+    it('should default to native Promise', function*() {
+      Nightmare.Promise.should.equal(Promise);
+      nightmare = Nightmare();
+      nightmare.should.be.ok;
+      var thenPromise = nightmare.goto('about:blank').then();
+      thenPromise.should.be.an.instanceof(Promise);
+      yield thenPromise;
+    });
+
+    it('should override default Promise library', function*() {
+      // `withDeprecationTracking()` messes w/ prototype constructor references
+      Nightmare.Promise = require('..').Promise = require('bluebird');
+      Nightmare.Promise.should.equal(require('bluebird'));
+      nightmare = Nightmare();
+      nightmare.should.be.ok;
+      var thenPromise = nightmare.goto('about:blank').then();
+      thenPromise.should.be.an.instanceof(require('bluebird'));
+      yield thenPromise;
+    });
+
+    it('should not override per-instance Promise library', function*() {
+      Nightmare.Promise.should.equal(Promise);
+      nightmare = Nightmare({ Promise: require('bluebird') });
+      nightmare.should.be.ok;
+      var thenPromise = nightmare.goto('about:blank').then();
+      thenPromise.should.not.be.an.instanceof(Promise);
+      thenPromise.should.be.an.instanceof(require('bluebird'));
+      yield thenPromise;
+    });
+  })
 
   describe('Nightmare.action(name, fn)', function() {
     afterEach(function*() {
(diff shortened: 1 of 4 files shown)


test/index.js
```diff
@@ -1892,8 +1892,62 @@ describe('Nightmare', function () {
       nightmare = Nightmare({ electronPath: require('electron') });
       nightmare.should.be.ok;
     });
+
+    it('should allow to use external Promise', function*() {
+      nightmare = Nightmare({ Promise: require('bluebird') });
```
jekku: maybe the require('bluebird') return can be assigned to a variable at the beginning of this generator function. Although I know that require is always cached, I wonder if it would look nicer if it's in a variable.

test/index.js
```diff
@@ -1892,8 +1892,62 @@ describe('Nightmare', function () {
...
+      nightmare.should.be.ok;
+      var thenPromise = nightmare.goto('about:blank').then();
```
jekku: maybe we can replace all these vars within this function with const since they are never mutated after.

test/index.js
```diff
@@ -1892,8 +1892,62 @@ describe('Nightmare', function () {
...
+      nightmare = Nightmare();
+      nightmare.should.be.ok;
+      var thenPromise = nightmare.goto('about:blank').then();
+      thenPromise.should.be.an.instanceof(Promise);
+      yield thenPromise;
+    });
+
+    it('should override default Promise library', function*() {
+      // `withDeprecationTracking()` messes w/ prototype constructor references
+      Nightmare.Promise = require('..').Promise = require('bluebird');
```
jekku: What's the difference of this test from the in the previous that checks if it should allow an external promise????

test/index.js
```diff
@@ -1892,8 +1892,62 @@ describe('Nightmare', function () {
...
+      Nightmare.Promise.should.equal(require('bluebird'));
+      nightmare = Nightmare();
+      nightmare.should.be.ok;
+      var thenPromise = nightmare.goto('about:blank').then();
+      thenPromise.should.be.an.instanceof(require('bluebird'));
+      yield thenPromise;
```
jekku: does it stop executing if yield is not called here?

itsjustcon: We need our promise routines to finish before our test ends, so there's 2 ways to give mocha our promise: return the promise to mocha at the end of the test, or yield it. Currently it seems `yield` is considered better than `return` b/c if/when the promise throws, it will throw an error w/ a call stack going back to that specific line in your test code, rather than mocha getting the promise, waiting for it to resolve, and handling the error reporting.

test/index.js
itsjustcon: There's a couple diff things going on there:
1. we want the promise library to *default to* global/native ES6 promises (if available)
2. we want to be able to **set** the *global* `Promise` factory for the entire Nightmare package. This means that from this point on, all new instances of Nightmare will use the given promise library: `Nightmare.Promise = require('bluebird')`
3. we want to set a *per-instance* `Promise` factory. This means that I just want this one Nightmare instance to use this promise library but the others shouldn't be affected:
```
const nightmare = new Nightmare();
nightmare.Promise = require('bluebird');
// nightmare.Promise is now Bluebird
// Nightmare.Promise is still what it was before
```

test/index.js
itsjustcon: The only reason I didn't do that is because people like myself who regularly use external Promise libs usually assign them to the `Promise` var in the beginning of their scripts. Doing so here overrides the global promise lib not just in the context of this function/test, but also globally (b/c `Promise` is defined globally w/ `var` & not `let`) so that will mess up other tests.

The other option is doing something like this, but I'm not a fan of how it looks by-comparison lol:
```js
const Promise2 = require('bluebird');
nightmare = Nightmare({ Promise: Promise2 });
```

test/index.js
itsjustcon: kk

test/index.js
itsjustcon: Also, the main way I'm using this is using ES6 object literals:
```js
import Nightmare from 'nightmare';
import Promise from 'bluebird';
// ...
const nightmare = new Nightmare({ Promise });
```

jekku: Thanks for teaching me that! :)

test/index.js
jekku: Got it :)

jekku:
LGTM :)

stephenmathieson:
Awesome work dude! Thanks!

PR #995, opened 2017-02-07
Bump electron version

itomas:


jekku:
Why though?

rosshinkley:
Wouldn't hurt to keep up on the Electron version, especially if the tests pass. I'd be open to this.

PR #1008, opened 2017-02-20
replace the 404 link with valid link

suniaoo:
previous link is 404, replace it with correct link address


stephenmathieson:
Nice, thanks!

PR #1009, opened 2017-02-22
Remove redundant docs for 'log' event from README

keithkml:
I noticed this issue while reading the docs


stephenmathieson:
Nice catch, thanks!

PR #1025, opened 2017-02-28
Improve confirm dialog behavior

rjanicek:
Enable confirm dialog to return default response or true if default response is undefined.
In browser, confirm dialog default selection is 'ok' so makes sense that true should be returned by default.


jakubchadim:
I really need this 👍  thanks

jakubchadim:
It seems to throw error from goto. If page is redirected  to anogether page from confirm right after load. Than goto fails and throw wait error after timeout.

casesandberg:
@jakubchadim is the goto silently failing still an issue, or was it resloved?

matthewmueller:
Going to close this since this PR is stale. Thanks for your efforts!

PR #1066, opened 2017-03-30
Expose electron's process memory API

asakusuma:
Would be nice to profile memory during headless test execution.

Essentially proxies to:
https://github.com/electron/electron/blob/master/docs/api/process.md#pro
cessgetprocessmemoryinfo


asakusuma:
Tried running circle ci twice and both times a different screenshot test timed out on a different version of node, so seems like the timeout needs to be increased.

asakusuma:
Not sure how this change would cause screenshoting to fail, only in node 6

```
1) Nightmare rendering should screenshot an up-to-date image of the page (1):

      AssertionError: expected [ 0, 138, 0 ] to deeply equal [ 0, 153, 0 ]
```

casesandberg:

PR #1068, opened 2017-03-31
Allow hidden DevTools

csabapalfi:
Is there a specific important reason for only allowing `openDevTools` when `show:true`?

This PR removes that restriction.

e.g. once this is merged [alexbardas/nightmare-har-plugin](https://github.com/alexbardas/nightmare-har-plugin) can be used with show:false to extract HAR for a page and all sort of other devtools functionality becomes available headless-ly.

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -192,7 +192,7 @@ var nightmare = Nightmare({
 ```
 
 ##### openDevTools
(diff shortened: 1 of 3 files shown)


csabapalfi:
Closing this as there's no activity for the past few months.

PR #1077, opened 2017-04-09
Set mouse position for mouse events

FredrikNoren:
This sets the mouse position of all the mouse events to be the center of the element that you interact with (currently it defaults to (0, 0)).


casesandberg:
Looks good to me! Can you write a test or two covering it?

PR #1081, opened 2017-04-10
Xpath support

deseman:
Added xpath support by adding `{ useXpath: true }` to the nightmare options.

In my first attempt I created a function to select an element by query selector or xpath, but I could not get that to work with the evaluate_now function which is why I added all the checks on the useXpath in every action which uses document.querySelector.


casesandberg:
I don't think this makes sense at this time as we are already changing so many things in the v3 implementation. Let's revisit this after that has been released.

PR #1122, opened 2017-05-02
fix: loginListener according to electron docs

boxtown:
Fixes login listener according to:
https://electron.atom.io/docs/api/web-contents/#event-login
https://electron.atom.io/docs/api/app/#event-login


boxtown:
Not sure why but auth seems to be working without the `.preventDefault()` so closing this PR

PR #1149, opened 2017-06-02
Update Readme.md

elderbas:
Worth it to have something that isn't really specific to Nightmare?

Since this ES6 feature is really usable in any library it seems a bit noisy to me to give an example of it being used here.

What do you think?


casesandberg:
I concur

PR #1162, opened 2017-06-19
Added option for enable/disable audio

agallardol:
Now you can enable/disable audio using webPreferences.webaudio property in Nightmare instance. Default webaudio was setted to false so nightmare has browser audio muted.

Instance option webPreferences.webaudio is automatically propagated to electron BrowserWindow instance, so if you want more information check https://github.com/electron/electron/blob/master/docs/api/browser-window.md


casesandberg:
LGTM, can you make sure the tests are passing and then I can merge this in!

casesandberg:

PR #1187, opened 2017-07-12
add double click event.

echinen:
- Add a new double-click event for anyone needing to use in some specific situation.


casesandberg:
This doesn't really fit in the library, so I am going to close this. Thank you for the work though!

PR #1215, opened 2017-08-04
add linting

roryrjb:
This PR adds linting to the repo. It is using [xo](https://github.com/sindresorhus/xo) and has been setup to match the dominant style already, rather than imposing a completely different style.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,6 +6,9 @@ CIRCLE_PROJECT_REPONAME ?= 1
 #set headless if not set
 HEADLESS ?= 1
 
+lint: node_modules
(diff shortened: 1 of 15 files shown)


casesandberg:
🙌

matthewmueller:
This is always a 🔥 topic with developers. I agree that we should add linting, but I think we should go with the most popular one. 

Right now the most popular ones are:

- xo: 2.5k downloads / day
- standard: 20k downloads / day
- prettier: 136k downloads / day

Let's go with prettier. I'll create an issue to move us over. Thanks though!

PR #1224, opened 2017-08-19
Some syntactic changes in Readme.md

jmporchet:


Readme.md
```diff
@@ -363,7 +363,7 @@ nightmare
 ```
 Note that callbacks support only one value argument (eg `function(err, value)`).  Ultimately, the callback will get wrapped in a native Promise and only be able to resolve a single value.
```
jmporchet: Is the word `value` really needed here?

casesandberg:
Great syntax changes!

PR #1234, opened 2017-08-23
Electron is unsafe to load malicious code in.

ejcx:
Because Nightmarejs is built on top of electron, it's possible
to execute arbitrary code if a malicious website is loaded.

Electron provides a README that explicitly states that the code
you load is your responsibility, and so Nightmare must inherit
this principle.


fouad:
Thanks @ejcx!

PR #1241, opened 2017-08-28
Added clickWithKeymap function to actions.js

ajbogh:
A feature of MouseEvents is to provide access to 4 different keybindings: shift, ctrl, alt, and media. This method exposes those bindings so that a test can perform a mouse click with one of the buttons down. 

The function is called similarly to the `click` event, except it allows a second parameter called `keyMap`. One example use case is if you first check a checkbox, then use `clickWithKeymap` to check the next checkbox while indicating that a key is pressed.

In the example below, the test checks the first checkbox, then clicks `cb3`, skipping `cb2`, however the `shiftKey` is set to `true` in the `keyMap`, so the client-side code will notice a shiftKey property exposed in the UI and will automatically check `cb2`. We use the test to verify that the client-side code acted appropriately.

I chose to implement this as a `click` event rather than a `check` event because it applies to other element types beyond checkboxes. It can apply to table rows for instance.

```
var checkboxes = yield nightmare
  .goto(fixture('manipulation'))
  .check('input[type=checkbox][name=cb1]')
  .clickWithKeymap('input[type=checkbox][name=cb3]', {shiftKey: true})
  .evaluate(function () {
    return document.querySelectorAll('input[type=checkbox]:checked');
  });
  Object.keys(checkboxes).length.should.equal(3);
```

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -291,6 +291,32 @@ Refresh the current page.
 #### .click(selector)
 Clicks the `selector` element once.
 
+#### .clickWithKeymap(selector, keyMap)
+Clicks the `selector` element once while pressing a key specified in the keymap. Keymap should contain the key to press with a true value.
+
+Available keys:
+```js
+{
+  shiftKey: true,
+  altKey: true,
+  ctrlKey: true,
+  mediaKey: true
+}
+```
+Example:
+```js
+nightmare
+  .goto(someUrl)
+  .click('input[type=checkbox].first')
+  .clickWithKeymap('input[type=checkbox].third', {shiftKey: true})
+  .evaluate(function () {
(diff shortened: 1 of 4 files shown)


ajbogh:
This code doesn't include the changes in #1077, but it's very similar.

matthewmueller:
Hi @ajbogh, thanks for taking the time to put this together and sharing your use case – I think this would be a bit better suited as a custom action outside of core.

For 3.x, we'll be looking to address #1311, which should cover this use case, particularly if we did something like this:

```js
await nightmare.keydown("leftShift")
await nightmare.click("#test")
await nightmare.keyup("leftShift")
```

Just like: https://github.com/GoogleChrome/puppeteer/blob/master/docs/api.md#class-keyboard

Issue #1060, opened 2017-03-27
! Massive security hole in nightmare


aight8:
The ```__nightmare``` object which is set on the window object, it contains the whole electron ipcRenderer on it, and it can be accessed by any website if they only want, every website have access to core electron features.

Furthermore just a deletion of the ```__nightmare``` (set it to null in the websites code) will freeze the evaluate method.

https://github.com/electron/electron/blob/master/docs/tutorial/security.md
https://github.com/electron/electron/issues/7929

chtrinh:
Maybe update the README.md with that link? 
I feel like this is a same concern as 'eval bash install scripts' you find online. 
Trust your source before you execute against it.

dyntopia:
@chtrinh most people probably wouldn't expect that a browser library
would allow a malicious site or a MITM to get RCE by simply asking the
crawler to execute commands through its IPC -- especially when the API
docs demonstrate how to crawl yahoo and other 3rd party sites over http

```sh
hji@nil:~/nightmare-ipc$ python3 exploit.py 172.16.1.1:8080
listening on 172.16.1.1:8080
172.16.1.2 - - [31/Mar/2017 21:11:37] "GET / HTTP/1.1" 200 -
=> we may have a shell on 172.16.1.2
$ id
uid=1003(crawl) gid=1003(crawl) groups=1003(crawl)
$ npm list nightmare
vuln@1.0.0 /home/crawl/vuln
└── nightmare@2.10.0
```
https://github.com/dyntopia/exploits/tree/master/nightmare-ipc

aight8:
This were my next point to write an exploit. Since it's public now how it can open a shell, it's just left the zero-day state.
The biggest problem is, it allow full machine access in worst case steal a whole companies data etc. Instead a hint I really recommend to write a big: do not use on untrusted sites. Until the ipcRenderer is in safe place.

ejcx:
Hi. I work on Security here at Segment and just heard of this issue. It's very interesting!

It is definitely not appropriate to allow websites to execute arbitrary code in a website that nightmare loads. We will have to fix this problem. I wonder if an easy fix is to somehow remove __nightmare from the global namespace. I am not super familiar with __nightmare, so I'm not sure if this is a small undertaking.

casesandberg:
Do you have some time to throw together a PR for this? I would be happy to take a look at it. I am going to close this, but feel free to open a new issue if you get a chance to work on it.

ejcx:
[...]

PR #1259, opened 2017-09-07
Sanitize what channels are accepted for ipc.send/sendSync

TimNZ:
Restrict channels for ipc messages from preload.js to ['console','response','error','log']

#1060 

Temp solution until decision made on longer term changes that might be breaking for people using custom preload


matthewmueller:
Awesome, I think we should just break it for 3.x and remove the potential vulnerability entirely.

matthewmueller:
This is done in #1358. Thanks for your efforts!

Issue #1269, opened 2017-09-22
installation fails on build server - failed to find Electron v1.8.0 for linux-x64


gmunisifreddy:
When we build one of the apps that uses nightmarejs on a linux CI server, i get the following. Any suggestions on how to solve the issue?

```
error /home/jenkins/workspace/xxx/node_modules/electron: Command failed.
Exit code: 1
Command: sh
Arguments: -c node install.js
Directory: /home/jenkins/workspace/monalisa/node_modules/electron
Output:
/home/jenkins/workspace/xxxx/node_modules/electron/install.js:48
  throw err
  ^

Error: Failed to find Electron v1.8.0 for linux-x64 at https://github.com/electron/electron/releases/download/v1.8.0/electron-v1.8.0-linux-x64.zip
    at Request.<anonymous> (/home/jenkins/workspace/xxxx/node_modules/nugget/index.js:169:61)
    at emitOne (events.js:96:13)
    at Request.emit (events.js:188:7)
    at Request.onRequestResponse (/home/jenkins/workspace/xxx/node_modules/request/request.js:1068:10)
    at emitOne (events.js:96:13)
    at ClientRequest.emit (events.js:188:7)
    at HTTPParser.parserOnIncomingClient [as onIncoming] (_http_client.js:473:21)
    at HTTPParser.parserOnHeadersComplete (_http_common.js:99:23)
    at TLSSocket.socketOnData (_http_client.js:362:20)
    at emitOne (events.js:96:13)
info Visit https://yarnpkg.com/en/docs/cli/add for documentation about this command.
{ Error: Command failed: yarn add node-server@0.6.3 appmeta-register-ebay@4.0.1 --dev --no-progress --no-emoji --ignore-engines --pure-lockfile --cache-folder /home/jenkins/workspace/monalisa/.yarn-cache
```

atkinchris:
This is due to Electron `v1.8.0` being deleted from GitHub, apparently due to a security vulnerability. (https://github.com/electron/electron/issues/10574#issuecomment-331373344)

Restricting Nightmare's dependency to an existing version would resolve this (or waiting till Electron publishes `v1.8.1` - but there's no ETA for this).

atkinchris:
I've raised https://github.com/segmentio/nightmare/pull/1270 to pin the version of Electron to the latest stable (`v1.7.6`).

matthewmueller:
Hey guys, it looks like the fix was merged in. If you're still running into problems, feel free to reopen!

PR #1270, opened 2017-09-22
Pin Electron version

atkinchris:
This pins the version of Electron used by the project, to prevent issues with the nuked Electron 1.8.0 version (https://github.com/electron/electron/issues/10574#issuecomment-331373344).

This addresses: https://github.com/segmentio/nightmare/issues/1269

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,7 +20,7 @@
     "debug": "^2.2.0",
     "deep-defaults": "^1.0.3",
     "defaults": "^1.0.2",
-    "electron": "^1.4.4",
+    "electron": "1.7.6",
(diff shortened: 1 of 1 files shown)


kanaabe:
Hi there, any idea as to if/when this might get merged? I'm trying to decide if I should uninstall nightmare and pend integration tests that use it or wait until this gets merged since it's currently holding up our CI/deploys. Thanks!

viko16:
Should we update now? 

`electron < 1.7.11` has a known potential security vulnerability on Windows.

via https://electronjs.org/blog/protocol-handler-fix

PR #1283, opened 2017-10-03
adding selector to the timeout error in .wait()

machadogj:
It is very useful to see which was the specific selector that caused the wait to timeout.


matthewmueller:
Hi @machadogj, thanks for the PR!

Do you mind cleaning up the PR a bit to just include the "adding selector to the timeout error in .wait()"?

The other commits could done as separate PRs (though I think we covered the -3 error code commit now)

machadogj:
Hi @matthewmueller, I’ll try to do it tomorrow. We’ve long switched to my fork, but it would be nice to go back to using the original package!

machadogj:
Closing in favor of #1381

PR #1339, opened 2017-12-08
Added ability to specify the client certificate selected by electron

zachary-sierakowski:
I could not find any simple way to access the `select-client-certificate` event on the electron instance via nightmare. This change will allow a nightmare user to select a client certificate by subjectName instead of always relying on the first available client certificate.


matthewmueller:
This looks great, thanks!

PR #1370, opened 2018-02-03
Use --save-dev instead of --save

JimTheMan:
This is a tests runner so I don't see why people would ever need this in the actual source when users are interacting with it. Won't people pretty much always use this as a dev dependency? I know webpack fixes it all, but people could still find it confusing. :)

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -54,7 +54,7 @@ nightmare
 You can run this with:
 
 ```shell
-npm install --save nightmare
+npm install --save-dev nightmare
 node example.js
 ```
 


airtoxin:
This is a __browser automation tool__ not a test runner.
There are may usages of this. crawler, debug, test runner is one of those, so no need to change readme.

f2prateek:
Agreed with @airtoxin! Your usage and requirements may vary, and it may be fine to use --save-dev or --save depending on the use case.

PR #1376, opened 2018-02-12
.wait() accepts multiple selectors as OR

mattxgreen:
Retains original functionality of .wait(), but now allows multiple selectors to be specified as an OR condition:

`.wait('div#this-one','div#that-one')`

This will wait for either selector. Optional soft timeout is still intact.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@ node_modules
 test/tmp
 test/DevTools Extensions
 .idea
+temp
--- a/lib/actions.js
+++ b/lib/actions.js
@@ -412,8 +412,8 @@ exports.wait = function () {
   var args = sliced(arguments);
   // var done = args[args.length-1];
   // change args in place to remove done()
-  var done = args.splice(-1,1);
-  if (args.length < 2) {
+  var done = args.splice(-1,1)[0];
+  if (args.length < 1) {
     debug('Not enough arguments for .wait()');
     return done();
   }
@@ -431,13 +431,11 @@ exports.wait = function () {
   }
   else if (typeof arg === 'string') {
     var timeout = null;
-    if (typeof args[args.length -1] === 'number') {
+    if (args.length > 1 && typeof args[args.length -1] === 'number') {
       timeout = args.splice(-1,1);
     }
-    args.map( selector => {
-      debug('.wait() for '+selector+' element'+(timeout ? ' or '+timeout+'msec' : ''));
-      waitelem.apply({ timeout: timeout }, [this, selector, done]);
-    })
+    debug('.wait() for '+args+' element(s)'+(timeout ? ' or '+timeout+'msec' : ''));
+    waitelem.apply({ timeout: timeout }, [this, ...args , done]);
   }
   else if (typeof arg === 'function') {
     debug('.wait() for fn');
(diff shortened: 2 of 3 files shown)


matthewmueller:
Hey @mattxgreen, thanks for taking the time to work on this!

I'm a bit torn on this because I personally think we've already overloaded `wait()` way too much. On the other hand I totally understand this use case.

I think it may make more sense to make this into a custom action and keep it out of core though. There's another use case I've found myself wanting around this which is wait until the first of N selectors becomes available. This is sort of the `Promise.all`, while there's another case for a `Promise.race`. But by changing the API in this way, we only cover one of these use cases. 

What do you think?

mattxgreen:
You'll have to elaborate on "the first of N selectors". I understand what you've stated to mean, when any one of a number of selectors is found. If that is the case, that is exactly what this PR does.

matthewmueller:
@mattxgreen oh my mistake, I had it reversed. I guess the opposite use case also seems valid to me: an `ALL` selectors condition. 

Given that both seem like equally valid use cases for a `.wait(sel1, sel2, sel3...)`, it could lead to some confusion. I wonder if we could implement `waitOne(...)` or `waitAll(...)`, though I still think it's probably best to just defer to custom actions.

mattxgreen:
I see both cases having merit, but for me this is the most useful case. I have been having troubles when navigating to a page where it can be one of two pages. .wait() with a fixed delay is unreliable, so being able to look for 'this' or 'that' is a reliable alternative. The most common use I see, is login success/fail, and waiting for either page. I'm also using it for a gradual rollout on ci/cd, and detecting the old version vs new version. I also see these cases being more common than a .wait() with multiple selectors for ALL, but I am obviously biased.

This PR is about 20 lines, very 'cheap', and doesn't modify any existing functionality. Would love to see it in the codebase, but if I'm the only one, I'll see if I can make it into an external module.

matthewmueller:
Let's keep it as a separate action for now – here's an example of how you'd set something like that up:

```js
const Nightmare = require('nightmare')

main().catch(console.error)

async function main() {
  Input(Nightmare)

  const nightmare = Nightmare({ show: true })

  await nightmare.goto('https://google.com')
  await nightmare.input('#lst-ib', 'nightmare is')
  await sleep(5000)
  await nightmare.input('#lst-ib', ' awesome')
  await nightmare.end()
}

function sleep(ms) {
  return new Promise(res => setTimeout(res, ms))
}

function Input(Nightmare) {
  Nightmare.action(
    'input',
      parent.respondTo('input', function(value, done) {
        var chars = String(value).split('')

        function type() {
          var ch = chars.shift()
          if (ch === undefined) {
            return done()
          }

          // keydown
          win.webContents.sendInputEvent({
            type: 'keyDown',
            keyCode: ch
          })

          // keypress
          win.webContents.sendInputEvent({
            type: 'char',
            keyCode: ch
          })

          // keyup
          win.webContents.sendInputEvent({
            type: 'keyUp',
            keyCode: ch
          })

          // defer function into next event loop
          setTimeout(type, options.typeInterval)
        }

        // start
        type()
      })
      done()
    },
    function(selector, text, done) {
      // focus, type
      return this.evaluate_now(
        selector => document.querySelector(selector).focus(),
        () => this.child.call('input', text, done),
        selector
      )
    }
  )
}
```

Thanks for your efforts!

PR #1381, opened 2018-02-19
adding selector to the timeout error in .wait()

machadogj:
It is very useful to see which was the specific selector that caused the wait to timeout.

(replacing #1283)


machadogj:
@matthewmueller here you go the isolated fix of #1283. Cheers!

matthewmueller:
thanks bud! need to look into why there's these failures, but they're unrelated to this change.

machadogj:
@matthewmueller you might want to take a look at this other one we had to patch to be able to use nightmare.js:

https://github.com/machadogj/nightmare/commit/cbc35ae2c15c824c9983d50b742e3363aa0d70a0

For some reason we kept on getting `navigation error` in our jenkins CI but screenshots didn't reveal anything wrong. Also, we were not able to reproduce it locally.

PR #1386, opened 2018-02-28
chore: add linting, formatting, consistent package management (via yarn) and a precommit hook

matthewmueller:
This PR modernizes the release of nightmare a bit and should cut down on some issues and PRs. 

Here's what's changed:

- Added the recommended linting via eslint (e.g. `eslint:recommended`)
- Added prettier formatting (prettier with single quotes and no semicolons)
- Install from an exact yarn version to avoid npm regressions like this (https://github.com/npm/npm/issues/17161)
- Added a git pre-commit hook via husky to run `make test` before each commit

Closes #1215

PR #1392, opened 2018-02-28
fix: node 4 by going back to npm

matthewmueller:
This PR takes us back to using NPM for our builds. This should fix node 4, which broke after #1386.


matthewmueller:
Hmm, nevermind. It must be something else. It works locally with node 4.7.2 and npm 2.15.11

PR #1454, opened 2018-05-25
toPng fix

wiliam-paradox:
In electron ^2.0.0 method image .toPng is called .toPNG
see: https://github.com/electron/electron/blob/master/docs/api/native-image.md#imagetopngoptions

### DIFF
--- a/lib/runner.js
+++ b/lib/runner.js
@@ -474,7 +474,16 @@ app.on('ready', function() {
     // https://gist.github.com/twolfson/0d374d9d7f26eefe7d38
     var args = [
       function handleCapture(img) {
-        done(null, img.toPng())
+        if(img.toPng) {
+            done(null, img.toPng());
+
+        } else if(img.toPNG) {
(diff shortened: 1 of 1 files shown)


kamthamc:
Can anyone please review this PR? its stopping from upgrading to newer versions of electron.

lib/runner.js
```diff
@@ -474,7 +474,16 @@ app.on('ready', function() {
...
-        done(null, img.toPng())
+        if(img.toPng) {
+            done(null, img.toPng());
+
+        } else if(img.toPNG) {
+            // @see https://github.com/electron/electron/blob/master/docs/api/native-image.md#imagetopngoptions
+            done(null, img.toPNG());
+
+        } else {
+            done(new Error('no png method found in electrons image.'));
```
wojtekmaj: I think that's what we have package.json for is to be sure this will never happen.

So, we could have it in one line:

```js
const toPng = img.toPng || img.toPNG;
done(null, toPng);
```
```

PR #1494, opened 2018-09-16
Commit lockfile

wojtekmaj:
> package-lock.json is automatically generated for any operations where npm modifies either the node_modules tree, or package.json. It describes the exact tree that was generated, such that subsequent installs are able to generate identical trees, regardless of intermediate dependency updates.

This is great because it prevents the "works on my machine" effect.


wojtekmaj:
Will need #1493 to be updated if merged first, and the other way around.

PR #1536, opened 2019-03-22
Use either toPng or toPNG when screenshot

IDisposable:
If the `img` objects being resolved in electron don't have a `toPng` method, they may have a `toPNG` method depending on the version of electron being used.

This change will use the correct one or resolve with a `ReferenceError`


IDisposable:
Would we prefer to use `Error` than `ReferenceError`?  Let me know and I'll adjust

Issue #1539, opened 2019-04-23
Will this project be updated for the latest version of Electron?


ashleydavis:
Nightmare is dependent on Electron 1.8.4.

The latest version of Electron is 4.1.4.

Will Nightmare be updated to the latest Electron?

honga:
+ 1 please update to the latest electron

netsgnut:
Now with #1534 merged, the latest version is 3.0.2 (as of writing), using an Electron version of 2.0.18, with Chrome 61 at its core.

For the long term development of the project, I believe it is inevitable to have an upgrade plan for Electron dependencies, e.g. a major version bump of Nightmare to track a major version bump of Electron. While I would love to fiddle around with the latest version of Electron myself as well (it seems now even v5.0.2 is out with Chrome 73), I guess it is OK too if we are to use v3 (Chrome 66) or v4 (Chrome 69).

Now given our current version is v3.0.2, perhaps from the next release, we could try to see if Nightmare v4 could track Electron v4 (e.g. 4.2)? It might be a good idea for Nightmare major version to be in sync with Electron major version, and Nightmare's own API changes (if small) and changing of Electron dependency within the same major branch (e.g. 4.1 -> 4.2) would bump a minor version.

In that case, existing apps would need to follow more closely to Electron's API breakages if they are to use some of Electron's APIs, though they can always be locking on a specific major version to avoid massive code changes.

Any ideas, @matthewmueller, and anything we the community could be of help? 

It is a pity that this project, while still super useful and powers a few of my pet projects, gets stuck in quite an old underlying engine. Currently, I am hacking a local copy of Nightmare to see if I could get it to work with Electron 4.2.2. I am still midway through and not quite sure of the implications of the architecture change underneath yet, but I will try my best there.

matthewmueller:
We'd happily accept a PR for bringing Electron up-to-date in Nightmare!

codebling:
Sorry for bumping an old issue. Not sure where to put this.

### Versioning

> a major version bump of Nightmare to track a major version bump of Electron

@netsgnut I disagree. Most users don't care what version of Electron is used by Nightmare under the hood, and shouldn't have to. I know that the subject is more complicated than that because some code has to be run directly by Electron. 

Tying Nightmare's version to Electron really complicates things and still seems like a bad idea. 

### Old Electron issues (nothing you haven't heard before)

[...]

PR #1598, opened 2020-03-27
Electron 8

codebling:
This is a work in progress. Opening PR for more visibility. 

# Current state 
Currently 10 tests are still failing. They are:

From `navigation`:

1) should succeed properly if request handler is present
2) should support javascript URLs that load pages
3) should fail immediately/not time out for 304 statuses

From `navigation/timeouts`:

4) should allow actions on a hanging page
5) should allow loading a new page after timing out

From `referer`:

6) should return referer from headers

From `options`:

7) should set a single header
8) should set all headers
9) should set headers for that request
10) should allow webPreferences settings

# Looking for help 

To help updating to Electron 8, submit PRs against the [branch here](https://github.com/codebling/nightmare/tree/electron-8). If there are issues, open them against that repo, as well.

# plan 

First step is to fix these tests. This is kind of a first "naive" pass at migrating. Since test coverage is quite good for this project, this might actually be all that is needed, but I'd still want to go over the [list of breaking changes](https://github.com/electron/electron/blob/b1f4ac00f09c75b39c775f20a8d07e308444c166/docs/breaking-changes.md) one by one to make sure each change has been accounted for.

### DIFF
--- a/lib/runner.js
+++ b/lib/runner.js
@@ -347,7 +347,7 @@ app.on('ready', function() {
         if (!protocol || KNOWN_PROTOCOLS.includes(protocol)) {
           return callback(true)
         }
-        electron.protocol.isProtocolHandled(protocol, callback)
+        callback(electron.protocol.isProtocolRegistered(protocol))
       }
 
       function startLoading() {
--- a/lib/runner.js
+++ b/lib/runner.js
@@ -508,15 +508,10 @@ app.on('ready', function() {
    */
 
   parent.respondTo('screenshot', function(path, clip, done) {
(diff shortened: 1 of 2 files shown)


codebling:
See also discussion in #1539

mjgp2:
@codebling - I was trying to look at your failing tests, but some of them I can't really get to the bottom of (e.g. javascript urls not working)

Did you make any more progress?

codebling:
@mjgp2 Thanks for looking! I didn't work on it any more! I feel like it's 70% there. I don't need it on the project I'm working on right now, so I probably won't look at this again until I do. I'll leave this parked here and hopefully someone can pick up where I left off.

PR #1671, opened 2024-04-20
Create datadog-synthetics.yml

ghost:


ghost:
This should 10x the speed of Segment through the Datadog JSON pipe

